PR #67, opened 2021-03-17
hw-mgmt: thermal: Indicate fan fault with led

toshic:
Set fan leds to green if there is no faults.
Indicate with orange led drawer with the faulty fan.

Signed-off-by: Anton Kortunov <toshik@yandex-team.ru>

### DIFF
--- a/usr/usr/bin/hw-management-thermal-control.sh
+++ b/usr/usr/bin/hw-management-thermal-control.sh
@@ -76,6 +76,7 @@ hw_management_path=/var/run/hw-management
 thermal_path=$hw_management_path/thermal
 config_path=$hw_management_path/config
 system_path=$hw_management_path/system
+led_path=$hw_management_path/led
 temp_fan_amb=$thermal_path/fan_amb
 temp_port_amb=$thermal_path/port_amb
 pwm=$thermal_path/pwm1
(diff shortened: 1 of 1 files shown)


usr/usr/bin/hw-management-thermal-control.sh
```diff
@@ -778,8 +779,26 @@ update_psu_fan_speed()
...
+set_fan_led()
+{
+	fan=$1
+	color=$2
+
+	if [ -f $led_path/led_fan"$fan"_$color ]; then
+		echo 255 > $led_path/led_fan"$fan"_$color
+	fi
+}
+
```
vadimp-nvidia, line 791: This is not good to manipulate with LED setting here.
Usually each OS provides it’s specific LED control process which is supposed to control all system LED according to the OS LED rules.
Doing setting here will break those processes, since only those process keep history of LED setting and in particular case of fan LED color could consider not only fault, but maybe also unexpected fan speed.
Code in hw-management is common for all OSes in which it’s running, so LED control should be provided in separate process. This process can provide the logic suggested in the pull request inside led control loop.

Also, there are some of hw-management users following the examples from https://github.com/Mellanox/hw-mgmt/tree/V.7.0010.2300_BR/examples/src/ev_hndl – using inotify watches for example for the case, when fan unit is inserted or removed.

PR #111, opened 2021-04-29
V.7.0010.3000 BR

mlnxmykolak:
- hw-mgmt: attributes: set delta 1100 limits
- hw-mgmt: attributes: add led trigger attribute


vadimp-nvidia (commented):
Approved.

PR #131, opened 2021-05-31
hw-mgmt: system: add msn3700 secured support

sholeksandr:

### DIFF
--- a/usr/usr/bin/hw-management.sh
+++ b/usr/usr/bin/hw-management.sh
@@ -195,6 +195,13 @@ mqm8700_rev1_base_connect_table=(    max11603 0x64 5 \
 			tmp102 0x4a 7 \
 			24c32 0x51 8)
 
+msn37xx_secured_connect_table=(    max11603 0x64 5 \
+			tps53679 0x70 5 \
+			tps53679 0x71 5 \
+			tmp102 0x49 7 \
+			tmp102 0x4a 7 \
+			24c512 0x51 8)
+
 msn3420_base_connect_table=(	max11603 0x6d 5 \
 			xdpe12284 0x62 5 \
 			xdpe12284 0x64 5 \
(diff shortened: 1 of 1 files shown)


sholeksandr:
RERUN

usr/usr/bin/hw-management.sh
```diff
@@ -603,6 +610,22 @@ mqmxxx_msn37x_msn34x_specific()
 	lm_sensors_config="$lm_sensors_configs_path/msn3700_sensors.conf"
 	get_i2c_bus_frequency_default
 }
+msn37xx_secured_specific()
```
vadimp-nvidia, line 613: What is the difference between msn37xx_secured_specific() and mqmxxx_msn37x_msn34x_specific()?

usr/usr/bin/hw-management.sh
```diff
@@ -195,6 +195,13 @@ mqm8700_rev1_base_connect_table=(    max11603 0x64 5 \
 			tmp102 0x4a 7 \
 			24c32 0x51 8)
 
+msn37xx_secured_connect_table=(    max11603 0x64 5 \
```
vadimp-nvidia, line 198: What is the difference between msn37xx_secured_connect_table() and mqm8700_base_connect_table()?

sholeksandr: Difference is in eeprom size: 24c512 0x51 8)

usr/usr/bin/hw-management.sh
sholeksandr: The difference  wit connection table with at24512 eeprom support
(${msn37xx_secured_connect_table[@]})

usr/usr/bin/hw-management.sh
vadimp-nvidia: Should we really add copy-paste for ~20 lines because of this one line change?
Maybe we can do it with parameter
Add
vpd_name=24c32

 to change in table to: "$vpd_name" 0x51 8

And set for relevant SKU:
vpd_name=24c512

sholeksandr: It will make code more difficult to read. I prefer to follow code structure which we already have - connection table per system type

vadimp-nvidia (approved):

sholeksandr:
RERUN

sholeksandr:
RERUN

PR #169, opened 2021-08-03
Fix issue of removing all temperature sensors links not related to switch. Delete just gearbox and module temperature links.

mshych:


mshych:
RERUN

mlnxmykolak (approved):

PR #180, opened 2021-08-16
hw-mgmt: debug: Add timeout to generate-dump.sh

sholeksandr:
Add timeout to generate-dump.sh

### DIFF
--- a/usr/usr/bin/hw-management-generate-dump.sh
+++ b/usr/usr/bin/hw-management-generate-dump.sh
@@ -42,13 +42,14 @@ MODE=$1
 dump_cmd () {
 	cmd=$1
 	output_fname=$2
+	timeout=$3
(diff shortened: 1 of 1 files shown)


sholeksandr:
RERUN

vadimp-nvidia (commented):
Sasha, why we need timeouts. Is it some case when we can't generate something and hang?

sholeksandr:
https://redmine.mellanox.com/issues/2754604
On ASIC FW stuck we have some commands hang in debug-dump like 'sensors', reading sfp, etc...

Best Regards,
Oleksandr Shamray
NVIDIA<http://www.nvidia.com/>

vadimp-nvidia (approved):

vadimp-nvidia:
RERUN

PR #185, opened 2021-08-20
hw-mgmt: chassis events: Add check if eeprom belong to mlxplat i2c bus

sholeksandr:
In order to not create hw-management links for eeprom which is out of
mlxplat i2c bus space add check for eeprom address.

It will prevent to add eeprom with addresses like:
/sys/devices/pci0000:00/0000:XX:XX.X/i2c-0/0-0050

### DIFF
--- a/usr/usr/bin/hw-management-chassis-events.sh
+++ b/usr/usr/bin/hw-management-chassis-events.sh
@@ -641,8 +641,11 @@ if [ "$1" == "add" ]; then
 				fi
 			fi
 		fi
-		ln -sf "$3""$4"/eeprom $eeprom_path/$eeprom_name 2>/dev/null
-		chmod 400 $eeprom_path/$eeprom_name 2>/dev/null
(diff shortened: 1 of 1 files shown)


usr/usr/bin/hw-management-chassis-events.sh
```diff
@@ -608,6 +608,9 @@ if [ "$1" == "add" ]; then
 	fi
 	if [ "$2" == "eeprom" ]; then
 		busdir="$3""$4"
+		if [[ $busdir != *"mlxplat"* ]]; then
```
vadimp-nvidia, line 611: Sssha, this logic is not so good.
It'll not work for DNI new system, which doesn't use mlx-platform driver.
We should find some more common criteria.
F.e. name of i2c driver.

vadimp-nvidia (changes requested):

vadimp-nvidia (approved):

PR #192, opened 2021-09-02
hw-mgmt: attributes: Fix cpld4 exposing on systems with 3 cpld

sholeksandr:
Fix cpld4 exposing on systems with 3 cpld
Add fan{x}_status existing check on fan_direction set
Set timeout for "journalctl" and "sx_sdk" on debug dump generation

### DIFF
--- a/usr/usr/bin/hw-management-start-post.sh
+++ b/usr/usr/bin/hw-management-start-post.sh
@@ -95,7 +95,7 @@ case $board in
 		;;
 esac
 
-timeout 20 bash -c 'until [ -L /var/run/hw-management/system/cpld1_version ]; do sleep 1; done'
+timeout 60 bash -c 'until [ -L /var/run/hw-management/system/cpld1_version ]; do sleep 1; done'
 sleep 1
 
 # Read cpld3 version with the mlxreg from mft package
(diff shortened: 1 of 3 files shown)


usr/usr/bin/hw-management-start-post.sh
```diff
@@ -95,7 +95,7 @@ case $board in
 		;;
 esac
 
-timeout 20 bash -c 'until [ -L /var/run/hw-management/system/cpld1_version ]; do sleep 1; done'
+timeout 60 bash -c 'until [ -L /var/run/hw-management/system/cpld1_version ]; do sleep 1; done'
```
vadimp-nvidia, line 98: We really could wait around 60 sec?
Even it affects only SPC1 systems, it could be some impact in Sonic, Cumulus,
Do you know what it could be so long?
Where you saw issue, in ONL or in other OS?

sholeksandr:
Hi Vadim.

On Liger we have delay more than 45 sec.

Best Regards,
Oleksandr Shamray
NVIDIA<http://www.nvidia.com/>

usr/usr/bin/hw-management-start-post.sh
sholeksandr: On MSN4600 we have delay more than 45 sec.

PR #196, opened 2021-09-13
hw-mgmt: psu_fw_update_tool fix proceed.

mlnxmykolak:
Change proceed flow. Fix msn4800 fail scenario.

Signed-off-by: Mykola Kostenok <c_mykolak@nvidia.com>

### DIFF
--- a/usr/usr/bin/hw_management_psu_fw_update_murata.py
+++ b/usr/usr/bin/hw_management_psu_fw_update_murata.py
@@ -325,19 +325,18 @@ def detect_address_60(i2c_bus, proceed):
     """
     i2c_detect = os.popen("i2cdetect -y {}".format(i2c_bus)).read()
     addr_60 = re.findall(r'60: (..)', i2c_detect)
-    if addr_60 and addr_60[0] == "60":
-        print("i2c address 0x60 ocupied.")
-        addr_70 = re.findall(r'70: (..)', i2c_detect)
-        if addr_70 and addr_70[0] == "70":
-            if not proceed:
+    if not proceed:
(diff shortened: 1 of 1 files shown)


mlnxmykolak:
RERUN

usr/usr/bin/hw_management_psu_fw_update_murata.py
```diff
@@ -325,19 +325,18 @@ def detect_address_60(i2c_bus, proceed):
...
     addr_60 = re.findall(r'60: (..)', i2c_detect)
-    if addr_60 and addr_60[0] == "60":
-        print("i2c address 0x60 ocupied.")
-        addr_70 = re.findall(r'70: (..)', i2c_detect)
-        if addr_70 and addr_70[0] == "70":
-            if not proceed:
+    if not proceed:
+        if addr_60 and addr_60[0] == "60":
+            print("i2c address 0x60 ocupied.")
+            addr_70 = re.findall(r'70: (..)', i2c_detect)
```
sholeksandr, line 331: Not good regexp. It will capture any 2 symbols (not only hex). Better to use something like this:
70: ([0-9a-f]{2})

mlnxmykolak:
RERUN

usr/usr/bin/hw_management_psu_fw_update_murata.py
mlnxmykolak: You right,  more accurate better than common ..
But this change not related to this regexp, its related to move "proceed" if above, to cover special flow on msn4800. 
Change regexp will need additional testing, so i prefer to do it in differ commit.

PR #208, opened 2021-09-27
hw-mgmt: chassis events: Fix iio name index for linecard

sholeksandr:
Set a2d ID value in sensor name "/lc/environment/a2d_iio:device{ID}_*"
to start from 0 for each linecard.


vadimp-nvidia (approved):

sholeksandr:
RERUN

PR #213, opened 2021-09-29
hw-mgmt: thermal: Use default dynamic minimum thermal profile for SN4800

vadimp-nvidia:
- Use default thermal profile 60% instead of 100%.
- Use temporary WA for PSU fan speed - keep 100% until PS firmware
  is not fixed.


vadimp-nvidia:
RERUN

sholeksandr (approved):

PR #218, opened 2021-10-11
V.7.0020.1100 BR

odmyko:
- hw-mgmt: system: new system SGN2410
- hw-mgmt: system: fix init config for sgn2410.


mlnxmykolak:
RERUN

PR #223, opened 2021-10-12
Fix Python scripts to be compatible with python2.7/python3 versions

sholeksandr:


sholeksandr:
RERUN

PR #228, opened 2021-10-14
hw-mgmt: patches: v4.19/4.9,5.10: Fix cpld*_pn register second byte reading

sholeksandr:
platform/x86: mlx-platform: Fix cpld*_pn register second byte reading

When we are reading register which is 2 bytes length we checking both bytes
for volatile/readable ability. We using a 2-byte type for reading CPLD{N}_PN
value. This fix adds a second (CPLD{N}_PN+1) register to the volatile/read
check function.


sholeksandr:
RERUN

PR #229, opened 2021-10-14
hw-mgmt: config: Fix list of kernel modules init

vadimp-nvidia:
Fix name of hotswap driver - it is called lm25066.


MellanoxBSP:
RERUN

sholeksandr:
RERUN

PR #243, opened 2021-11-03
hw-mgmt: attributes: Fix PSU power sensor attributes capability

sholeksandr:
Fix PSU power sensor attributes capability crating. Add empty capability file crating
even if sensor haven't additional attributes.


sholeksandr:
RERUN

PR #244, opened 2021-11-04
hw-mgmt: attributes: Fix cpld3 version reading on spc1 systems

sholeksandr:
Fix cpld3 version reading on spc1 systems


sholeksandr:
RERUN

PR #256, opened 2021-11-08
V.7.0020.1200 BR

vadimp-nvidia:
- hw-mgmt: patches: 4.19: Add squash patch for MTECR register
- hw-mgmt: patches: 4.19: Rename patch and add to table list


MellanoxBSP:
RERUN

PR #258, opened 2021-11-09
hw-mgmt: system: Disable ASIC oprations for systems that not supporting it

sholeksandr:
Disabled ASIC operations for systems that do not support it. Affected SKU: HI132


sholeksandr:
RERUN

PR #263, opened 2021-11-11
hw-mgmt: chassis events: Add voltmon connection for specific system type

sholeksandr:
In some cases we can have duplicated of i2c bus/addr combination for several
different systems. In this case we will have duplication in udev rules and
several rules will be called twice. Second rule call of the same device can
overwrite already created links for voltmon. This fix add check for specific
combination of system type and voltmon address.


MellanoxBSP (approved):

sholeksandr:
RERUN

PR #275, opened 2021-11-24
hw-mgmt: patches: 4.19: Fix asic[1/2]_reset control Fix asic[1/2]_reset control

sholeksandr:


sholeksandr:
RERUN

PR #280, opened 2021-12-01
hw-mgmt: thermal: Add sanity check for ASIC temperature

vadimp-nvidia:
In some marginal cases ASIC temperature can grow up too fast.
Along with it at this time user space thermal loop could decrease fan
speed according to dynamic minimum value. If it happens at the
beginning of ASIC warming up trend, it can cause kernel's cooling
control algorithm stacking in dormant state, in which kernel will stop
fan cooling level increasing. It is extremely rare scenario, but it
worth to add protection against it.


vadimp-nvidia:
RERUN

vadimp-nvidia:
RERUN

vadimp-nvidia:
RERUN

vadimp-nvidia:
RERUN

sholeksandr (approved):

PR #283, opened 2021-12-08
hw-mgmt: udev: Fix fan_amb matching rule for SN2201.

mshych:


MellanoxBSP:
RERUN

PR #287, opened 2021-12-13
hw-mgmt: rules: Replace QSFP rules by one generic rule

vadimp-nvidia:
Use generic rule for 'netdevice' renaming.


sholeksandr:
RERUN

vadimp-nvidia:
RERUN

PR #289, opened 2021-12-15
hw-mgmt: rules: Add generic rule for netdevice renaming

vadimp-nvidia:
Add generic rule for netdevice. This rule is supposed to work
with kernel 5.10, becasue driver has support for devlink
callback, which activates newly added rule.
Old and new rules should not impact each other.


vadimp-nvidia:
RERUN

felixradensky (approved):

felixradensky:
RERUN

PR #290, opened 2021-12-15
hw-mgmt: kernel pacthes: 4.19: Modify patch #0134

vadimp-nvidia:
Patch #0134 is modifed due to following issue.
When current cooling level is set from user space to 10 + 'n' for
configuration purpose, due to return of non-zero value, this request
is retried by user space and it is resulted in the below flow:

cat /var/run/hw-management/thermal/cooling_cur_state
8
echo 16 > /var/run/hw-management/thermal/cooling_cur_state
cat /var/run/hw-management/thermal/cooling_cur_state
6

While 16 should just set limit and not change level from 8 to 6.


vadimp-nvidia:
RERUN

felixradensky (approved):

PR #292, opened 2021-12-16
hw-mgmt: patches 4.19/5.10: Align patches with upstream add new

vadimp-nvidia:
Align several pacthes with upstream code.
Add new pacthes accepted for v5.16.


vadimp-nvidia:
RERUN

recipes-kernel/linux/linux-5.10/0052-hwmon-mlxreg-fan-Return-non-zero-value-when-fan-curr.patch
```diff
@@ -68,47 +68,50 @@ Here is the trace from KASAN:
...
  	 * overwritten.
  	 */
  	if (state >= MLXREG_FAN_SPEED_MIN && state <= MLXREG_FAN_SPEED_MAX) {
-+		config = true;
++		/*
++		 * This is configuration change, which is only supported through sysfs.
++		 * For configuration non-zero value is to be returned to avoid thermal
++		 * statistics update.
++		 */
++		config = 0; /*1*/;
```
sholeksandr, line 96: What the sense of this code?

config = true;
config = 0;

vadimp-nvidia:
config = 1 was added to indicate this is configuration action and to not update thermal statistcics.
However this pacth is not good.

Same as I did in:

    hw-mgmt: kernel pacthes: 4.19: Modify patch #0134

    Patch #0134 is modifed due to following issue.
    When current cooling level is set from user space to 10 + 'n' for
    configuration purpose, due to return of non-zero value, this request
    is retried by user space and it is resulted in the below flow:

    cat /var/run/hw-management/thermal/cooling_cur_state
    8
    echo 16 > /var/run/hw-management/thermal/cooling_cur_state
    cat /var/run/hw-management/thermal/cooling_cur_state
    6

    While 16 should just set limit and not change level from 8 to 6.

    Signed-off-by: Vadim Pasternak ***@***.******@***.***>

recipes-kernel/linux/linux-5.10/0081-platform-x86-mlx-platform-Fix-cpld-_pn-register-seco.patch
```diff
@@ -1,66 +0,0 @@
-From ec25e4c96f6fce254574893c2d56c43fee8d3331 Mon Sep 17 00:00:00 2001
```
sholeksandr, line 1: This is fix for PN reading. Why this was deleted?

vadimp-nvidia:
I squashed it to 0056-platform-x86-mlx-platform-Add-initial-support-for-ne.patch

sholeksandr (approved):

PR #298, opened 2021-12-21
hw-mgmt: rules: Add validate of port name attribute

vadimp-nvidia:
In order to avoid conflicts for ‘netdevice’ renaming, add validation of
port name attribute. Apply generic ‘udev’ rule only in case this
attribute is not empty.


vadimp-nvidia:
RERUN

sholeksandr (approved):

mshych (approved):

PR #301, opened 2021-12-23
hw-mgmt: scripts: workaround for weird CPU cores numbering on SN2201

mshych:
Intel CPU coretemp driver represents Denverton CPU cores as number 6 and 12
instead 0, 1. Make hw-mgmt links with usual core0, core1 names.


mshych:
RERUN

felixradensky (approved):

PR #306, opened 2022-01-04
hw-mgmt: psu fw update tool: add show FW version command

sholeksandr:
Add argument to show PSU FW version in PSU FW update scripts

### DIFF
--- a/usr/usr/bin/hw_management_psu_fw_update_delta.py
+++ b/usr/usr/bin/hw_management_psu_fw_update_delta.py
@@ -50,6 +50,7 @@ import os
 
 import hw_management_psu_fw_update_common as psu_upd_cmn
 
+TOOL_VERSION = '1.0'
 MFR_FWUPLOAD_MODE = 0xd6
 MFR_FWUPLOAD = 0xd7
 MFR_FWUPLOAD_STATUS = 0xd2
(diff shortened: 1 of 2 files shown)


usr/usr/bin/hw_management_psu_fw_update_delta.py
```diff
@@ -50,6 +50,7 @@
 
 import hw_management_psu_fw_update_common as psu_upd_cmn
 
+TOOL_VERSION = '1.0'
```
vadimp-nvidia, line 53: Sasha, I suppose we should also expose PSU FW version through
cat /var/run/hw-management/eeprom/psu{n}_vpd by adding FW version field at bottom of the below?
MFR_NAME: Murata-PS
CHSUM_FIELD: 0x81 0x9e
PN_VPD_FIELD: MTEF-PSF-AC-L
SN_VPD_FIELD: MT2118X09951
REV_VPD_FIELD: A1
MFG_DATE_FIELD: 12/05/21
CAPACITY: 2000
MIN_RPM: 0
MAX_RPM: 24000
FEED: AC
CALC_CRC: 0x81 0x9e
FW Version: <>

vadimp-nvidia (approved):

usr/usr/bin/hw_management_psu_fw_update_delta.py
sholeksandr: PSU VPD contents taking form vendor specific area of PMBUS commands. It doesn't have FW version.
I don't think that we need to expose FW version here.

sholeksandr:
RERUN

PR #310, opened 2022-01-10
hw-mgmt: doc: kernel config clarifications. Update pkg version.

mshych:


mshych:
RERUN

PR #312, opened 2022-01-10
hw-mgmt: attributes: Add PSU FW version attributes

felixradensky:

### DIFF
--- a/usr/usr/bin/hw-management-thermal-events.sh
+++ b/usr/usr/bin/hw-management-thermal-events.sh
@@ -657,6 +657,21 @@ if [ "$1" == "add" ]; then
 			echo $ps_min_rpm > "$thermal_path"/"$2"_fan_min
 		fi
 
+		# PSU FW VER
+		mfr=$(grep MFR_NAME $eeprom_path/"$2"_vpd | awk '{print $2}')
+		if echo $mfr | grep -iq "Murata"; then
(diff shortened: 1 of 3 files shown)


usr/usr/bin/hw-management-thermal-events.sh
```diff
@@ -657,6 +657,21 @@ if [ "$1" == "add" ]; then
...
+		# PSU FW VER
+		mfr=$(grep MFR_NAME $eeprom_path/"$2"_vpd | awk '{print $2}')
+		if echo $mfr | grep -iq "Murata"; then
+			ps_ver=$(hw_management_psu_fw_update_murata.py -b $bus -a $psu_addr -v |
+				grep PSU | awk '{print $3}' | awk -F: '{print $2}')
+			echo $ps_ver > $fw_path/"$2"_fw
+			ps_ver=$(hw_management_psu_fw_update_murata.py -P -b $bus -a $psu_addr -v |
+				grep PSU | awk '{print $3}' | awk -F: '{print $2}')
+			echo $ps_ver > $fw_path/"$2"_fw_primary
+		elif echo $mfr | grep -iq "Delta"; then
```
vadimp-nvidia, line 669: We need to check if we don't enter Delta case for Alligator. If yes, we should filter it out by VMOD validation.

PR #316, opened 2022-01-12
hw-mgmt: psu fw update tool: Simpify FW version reporting output

felixradensky:
Report only PSU FW version. Remove reporting of tool version, bus,
address and MCU type to make the output easier to read and parse.

### DIFF
--- a/usr/usr/bin/hw-management-thermal-events.sh
+++ b/usr/usr/bin/hw-management-thermal-events.sh
@@ -660,18 +660,12 @@ if [ "$1" == "add" ]; then
 		# PSU FW VER
 		mfr=$(grep MFR_NAME $eeprom_path/"$2"_vpd | awk '{print $2}')
 		if echo $mfr | grep -iq "Murata"; then
-			ps_ver=$(hw_management_psu_fw_update_murata.py -b $bus -a $psu_addr -v |
-				grep PSU | awk '{print $3}' | awk -F: '{print $2}')
-			echo $ps_ver > $fw_path/"$2"_fw
-			ps_ver=$(hw_management_psu_fw_update_murata.py -P -b $bus -a $psu_addr -v |
-				grep PSU | awk '{print $3}' | awk -F: '{print $2}')
-			echo $ps_ver > $fw_path/"$2"_fw_primary
+			hw_management_psu_fw_update_murata.py -v -b $bus -a $psu_addr > $fw_path/"$2"_fw
+			hw_management_psu_fw_update_murata.py -v -b $bus -a $psu_addr -P > $fw_path/"$2"_fw_primary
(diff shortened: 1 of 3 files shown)


usr/usr/bin/hw_management_psu_fw_update_murata.py
```diff
@@ -379,12 +379,8 @@ def detect_address_60(i2c_bus, proceed):
     # bootloader_status(i2c_bus, i2c_adr)
     # burn_fw_file(i2c_bus, i2c_addr)
     if args.version:
-        print("Murrata FW update tool version:{}".format(TOOL_VERSION))
         fw_rev = read_murata_fw_revision(args.i2c_bus, args.i2c_addr, args.primary)
-        print("PSU FW version:{} (BUS:{}, Addr:{}, primary:{})".format(fw_rev,
-                                                                       args.i2c_bus, 
-                                                                       args.i2c_addr,
-                                                                       args.primary))
+        print(fw_rev)
```
vadimp-nvidia, line 383: Yes, I agree, we should also remove "PSU FW version:", since we have self-explaining attribute name.

usr/usr/bin/hw-management-thermal-events.sh
```diff
@@ -660,18 +660,12 @@ if [ "$1" == "add" ]; then
...
-				grep PSU | awk '{print $3}' | awk -F: '{print $2}')
-			echo $ps_ver > $fw_path/"$2"_fw
-			ps_ver=$(hw_management_psu_fw_update_murata.py -P -b $bus -a $psu_addr -v |
-				grep PSU | awk '{print $3}' | awk -F: '{print $2}')
-			echo $ps_ver > $fw_path/"$2"_fw_primary
+			hw_management_psu_fw_update_murata.py -v -b $bus -a $psu_addr > $fw_path/"$2"_fw
+			hw_management_psu_fw_update_murata.py -v -b $bus -a $psu_addr -P > $fw_path/"$2"_fw_primary
 		elif echo $mfr | grep -iq "Delta"; then
 			# Skip SN2201, FW update is not supported on this system
 			if [ "$board_type" != "VMOD0014" ]; then
```
vadimp-nvidia, line 667: Here, I think we'll need to extend our "blacklist". When we'll do it, maybe better to use "swicth" "case" instead "if"

vadimp-nvidia (approved):

felixradensky:
RERUN

usr/usr/bin/hw_management_psu_fw_update_murata.py
felixradensky: "PSU FW version" was also removed by this patch, only version itself is printed now.

PR #321, opened 2022-01-17
hw-mgmt: psu fw update tool: Allow skipping PSU redundancy check

felixradensky:
Add -S command line parameter to support skipping PSU redundancy check.
This makes it possible to update PSU FW on systems with a single active
PSU. This option already exists for Murata PSUs and is used for primary
and secondary FW update. Add it for Delta PSUs as well.

With PSU redundancy check disabled, Murata script will not perform PSU
reset after FW update, as this will cause power cycle on systems with
single active PSU.


sholeksandr (approved):

felixradensky:
RERUN

PR #322, opened 2022-01-18
hw-mgmt: thermal: add actual TC defines for SN2201.

mshych:


mshych:
RERUN

sholeksandr (approved):

PR #326, opened 2022-01-19
hw-mgmt: kernel 5.10: change patch 0089 of nvsw-sn2201 driver.

mshych:
Fix power-cycle on SN2201.


mshych:
RERUN

felixradensky (approved):

PR #327, opened 2022-01-20
hw-mgmt: system: Speed up ASIC de-initialization for Spectrum1/2 systems

vadimp-nvidia:
Speed up only ASIC de-initialization by setting I2C half cycle register to
400KHz before probing and setting it back to original value in CPLD
after probing. It should switch to 400KHz only for ASIC I2C driver
removing (for probing already it is already supported) to reduce it
time in order to avoid collision with chip reset coming from SDK.

Apply this functionality only for specific system types for both
probing and removing flows.


vadimp-nvidia:
RERUN

vadimp-nvidia:
RERUN

PR #331, opened 2022-01-24
V.7.0020.2000 BR

mshych:
- hw-mgmt: version: set version 7.0020.1334
- hw-mgmt: script & kernels: Change FAN hotplug event on SN2201 system


mshych:
RERUN

PR #333, opened 2022-01-25
hw-mgmt: script & udev: Add NVME temperature reports.

mshych:
Add NVME SSD temperature reports to hw-management.

### DIFF
--- a/usr/lib/udev/rules.d/50-hw-management-events.rules
+++ b/usr/lib/udev/rules.d/50-hw-management-events.rules
@@ -80,6 +80,9 @@ SUBSYSTEM=="hwmon", DEVPATH=="/devices/virtual/thermal/thermal_zone*/hwmon*", AC
 # SODIMM temp
 SUBSYSTEM=="hwmon", DEVPATH=="/devices/pci0000:00/*/i2c-*/0-001*/hwmon/hwmon*", ACTION=="add", RUN+="/usr/bin/hw-management-thermal-events.sh add sodimm_temp %p %k %S %n"
 
+# NVME temp
+SUBSYSTEM=="hwmon", DEVPATH=="/devices/pci0000:*/0000:*:*.*/*:*:*.*/hwmon/hwmon*", ACTION=="add", RUN+="/usr/bin/hw-management-thermal-events.sh add nvme_temp %S %p"
+
 # Switch - FAN tachometers, ASIC and ports temperatures and faults (I2C).
 SUBSYSTEM=="hwmon", DEVPATH=="/devices/platform/mlxplat/i2c_mlxcpld*/i2c-*/i2c-*/*-0048/hwmon/hwmon*", ACTION=="add", RUN+="/usr/bin/hw-management-thermal-events.sh add switch %S %p"
 SUBSYSTEM=="hwmon", DEVPATH=="/devices/platform/mlxplat/i2c_mlxcpld*/i2c-*/i2c-*/*-0048/hwmon/hwmon*", ACTION=="remove", RUN+="/usr/bin/hw-management-thermal-events.sh rm switch %S %p"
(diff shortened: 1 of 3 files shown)


mshych:
RERUN

usr/lib/udev/rules.d/50-hw-management-events.rules
```diff
@@ -80,6 +80,9 @@ SUBSYSTEM=="hwmon", DEVPATH=="/devices/virtual/thermal/thermal_zone*/hwmon*", AC
 # SODIMM temp
 SUBSYSTEM=="hwmon", DEVPATH=="/devices/pci0000:00/*/i2c-*/0-001*/hwmon/hwmon*", ACTION=="add", RUN+="/usr/bin/hw-management-thermal-events.sh add sodimm_temp %p %k %S %n"
 
+# NVME temp
+SUBSYSTEM=="hwmon", DEVPATH=="/devices/pci0000:*/0000:*:*.*/*:*:*.*/hwmon/hwmon*", ACTION=="add", RUN+="/usr/bin/hw-management-thermal-events.sh add nvme_temp %S %p"
```
sholeksandr, line 84: This DEVPATH mask not accurate. The /devices/pci0000:*/0000:*:*.*/*:*:*.*/ can fit many devices.

usr/usr/bin/hw-management-thermal-events.sh
```diff
@@ -687,6 +687,26 @@ if [ "$1" == "add" ]; then
...
 		/usr/bin/hw-management.sh chipup
 	fi
+	if [ "$2" == "nvme_temp" ]; then
+		dev_name=$(cat "$3""$4"/name)
+		if [ "$dev_name" == "nvme" ]; then
+			for i in {1..4}; do
+				if [ -f "$3""$4"/temp"$i"_input ]; then
+					label=$(cat "$3""$4"/temp"$i"_label | awk '{ gsub (" ", "", $0); print}')
+					name=$(echo "$label" | awk '{print tolower($0)}')
+					ln -sf "$3""$4"/temp"$i"_input "$thermal_path"/"$dev_name"_"$name"
```
sholeksandr, line 697: To create links you can use more safe function: check_n_link instead of "ln -s"

usr/lib/udev/rules.d/50-hw-management-events.rules
mshych: This is a common path to PCIe ports. NVME or not hwmon is checked inside the handler.
This was done intentionally. 
We can put a specific patch but in this way, it should be always updated.
I don't think that we will have many PCIe hwmon devices.

usr/usr/bin/hw-management-thermal-events.sh
mshych: check_n_link() checks if link exist or not.
There is a common if the main link to temperature exist: "if [ -f "$3""$4"/temp"$i"_input ]; then "
All other links are created in the same case thus I'm not sure that we should check the existence of every link 
in the same case.

mshych:
To summarize:
1. I can put the exact paths of NVME SSD of SN2201 and Vulcan carrier board, 2 current systems with NVMe.
In this case, we will not receive extra udev events (not so much) but should always check and extend in the future.
I.e. use
/sys/devices/pci0000:00/0000:00:09.0/0000:02:00.0/hwmon/hwmon*
/sys/devices/pci0000:00/0000:00:01.1/0000:02:00.0/hwmon/hwmon*
?
2. I don't see a real reason to use check_n_link in this case. Additional links can't exist without the main link to temp_input 
that is already checked.

mshych:
Yes, DRIVERS=="nvme" works together with PCIe wildcard address. I'll change udev rules.

mshych:
RERUN

vadimp-nvidia (approved):

PR #336, opened 2022-01-27
hw-mgmt: patches: 5.10: Add support for new system XH3000

vadimp-nvidia:
Add support for new system type XH3000, which is a water cooling
Ethernet switch blade equipped with 32x200G Ethernet ports.

The system is recognized by "DMI_BOARD_NAME" and "DMI_PRODUCT_SKU"
matches, when these fields are set to "VMOD0005" and "HI139"
respectively.


vadimp-nvidia:
RERUN

sholeksandr (approved):

PR #344, opened 2022-01-31
hw-mgmt: infra: Add support for mapping PCI to I2C ASIC bus

vadimp-nvidia:
Add dynamic mapping between PCIe and I2C ASIC busses.
Support only for new coming systems.

The following attributes will be available:
/var/run/hw-management/config/
	asic_num
	asic1_i2c_bus_id
	asic2_i2c_bus_id (for systems with two ASICs)
        asic1_pci_bus_id (for new systems)
	asic2_pci_bus_id (for new systems with two ASICs)

Signed-off-by: Vadim Pasternak <vadimp@nvdia.com>


vadimp-nvidia:
RERUN

mshych (approved):

PR #160, opened 2021-07-26
hw-mgmt: system: Change fru eeprom type to 24c512 for mqm9700 rev1 Change fru eeprom type 24c32->24c512 for mqm9700 rev1

sholeksandr:
Signed-off-by: Oleksandr Shamray <oleksandrs@nvidia.com>

PR #349, opened 2022-02-02
hw-mgmg: patches: 5.10: Update register map

vadimp-nvidia:
Update register map in patch #0157.
Rebase follwing after patches: #0158 - #0160.


vadimp-nvidia:
RERUN

sholeksandr (approved):

PR #354, opened 2022-02-15
hw-mgmt: scripts: Fix FAN EEPROM insert/removal on SN2201. 		  Common i2c_bus_offset fix.

mshych:
1. Use common helper function for connect/disconenct FAN EEPROMs on SN2201.
2. Fix common usage of i2c_bus_offset. New i2c_bus_offset attribute is added
   to hw-management config directory.

### DIFF
--- a/usr/usr/bin/hw-management-helpers.sh
+++ b/usr/usr/bin/hw-management-helpers.sh
@@ -114,12 +114,17 @@ find_i2c_bus()
     # Find physical bus number of Mellanox I2C controller. The default
     # number is 1, but it could be assigned to others id numbers on
     # systems with different CPU types.
+    if [ -f $config_path/i2c_bus_offset ]; then
+        i2c_bus_offset=$(< $config_path/i2c_bus_offset)
+        return
+    fi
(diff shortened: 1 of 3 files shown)


mshych:
RERUN

felixradensky (approved):

usr/usr/bin/hw-management-helpers.sh
```diff
@@ -182,12 +187,14 @@ change_file_counter()
 
 connect_device()
 {
+	find_i2c_bus
 	if [ -f /sys/bus/i2c/devices/i2c-"$3"/new_device ]; then
 		addr=$(echo "$2" | tail -c +3)
```
vadimp-nvidia, line 192: Why do we need to find addr like addr=$(echo "$2" | tail -c +3)? Before we just pass it as argument.

mshych: It was added by Sasha in previous commits. I can just say that this working.

mshych:
RERUN

sholeksandr (approved):

PR #358, opened 2022-02-15
hw-mgmt: topolgy: Add EROT events support for p4697 system

sholeksandr:
Add EROT events support for p4697 system

### DIFF
--- a/usr/usr/bin/hw-management.sh
+++ b/usr/usr/bin/hw-management.sh
@@ -77,6 +77,7 @@ hotplug_psus=2
 hotplug_fans=6
 hotplug_pwrs=2
 hotplug_linecards=0
+erot_count=0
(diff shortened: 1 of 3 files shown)


usr/usr/bin/hw-management.sh
```diff
@@ -77,6 +77,7 @@ hotplug_psus=2
 hotplug_fans=6
 hotplug_pwrs=2
 hotplug_linecards=0
+hotplug_erots=0
```
vadimp-nvidia, line 80: This is actually is not "hotplug". Maybe be this case it's better to call it here differently? Just erots_num f.r.?

vadimp-nvidia (approved):

sholeksandr:
RERUN

vadimp-nvidia:
RERUN

PR #372, opened 2022-03-08
kernel: patches: 4.19/5.10: Add SPI cnhl select for NVLink managed switch

sholeksandr:


sholeksandr:
RERUN

felixradensky (commented):
There's a typo in the subject, should be chnl/channel instead of cnhl

vadimp-nvidia (approved):

PR #377, opened 2022-03-23
hw-mgmt: events: Fix handling of CPLD3 version on SPC1 systems

felixradensky:
Disable removal of CPLD3 version file during chipdown.
This fixes the problem of this file missing after chipdown-chip
sequence. Removing this file is unnecessary as CPLD version
cannot change without power cycle.


vadimp-nvidia (approved):

sholeksandr:
RERUN

PR #386, opened 2022-04-03
hw-mgmt: scripts: fix mix-up of jtag tdi - tdo pins on SN2201

mshych:


mshych:
RERUN

vadimp-nvidia (approved):

PR #388, opened 2022-04-05
Create github action for adding release assetrs.

sholeksandr:
Create github action for adding release assetrs.


ciju-nvidia (approved):
Would be good to add a  bit more details in the 'Release' task.
I see that v3 scripts are available. Would you want to try those instead of v2?

PR #390, opened 2022-04-10
hw-mgmt: patches: 4.19/5.10 Remove status led fan from sysfs on SN2201 system.

mshych:
Don't export status led fan on front panel in sysfs.
There is no such API on all our other systems thus remove it also on SN2201.
This led shoudn't be handled by OS (Sonic, Cumulus).


mshych:
RERUN

vadimp-nvidia (approved):

mshych:
RERUN

mshych:
RERUN

PR #392, opened 2022-04-11
hw-mgmt: version: update package version to 7.0020.2038

mshych:


mshych:
RERUN

PR #397, opened 2022-04-18
hw-mgmt: patches: 4.19/5.10 Add FAN LEDs registers to SN2201 system.

mshych:
Add "dummy" CPLD FAN LEDs registers to the SN2201 system.
CPLD doesn't handle FAN LEDs on SN2201 but this register will allow triggering of actual LED change
through GPIO by hw-mgmt Udev handler script.
CPLD will change the state of the front panel FAN LED according to these registers.
Also, these registers add alignment of hw-mgmt attributes on SN2201 with other systems.


mshych:
RERUN

vadimp-nvidia (approved):

PR #401, opened 2022-04-25
hw-mgmt: patches: 4.19/5.10 Fix LED num for NVLink4 managed switch

sholeksandr:
Fix LED num for NVLink4 managed switch.


sholeksandr:
RERUN

PR #405, opened 2022-04-26
hw-mgmt: topology: Fix ASIC chip up/down command

sholeksandr:
This commit fixed flow when we try to do chip up/down on the ASIC's that not
supporting this feature.

S commit fix flow when we try to do chip up/down on ASIC that no


sholeksandr:
RERUN

vadimp-nvidia (approved):

sholeksandr:
RERUN

PR #407, opened 2022-04-27
hw-mgmt: patches: 4.19/5.10: Add sending udev event from leds-mlxreg driver.

mshych:
Add sending udev event from leds-mlxreg driver in case of color change.
Likely these patches will not be accepted upstream as they provide not
so regular flow.
Patches should be taken to NOS's as exceptions in upstream.


mshych:
RERUN

felixradensky (approved):

PR #408, opened 2022-04-27
hw-mgmt: topology: Fix sensors address for SGX H100 system Fix i2c sensors address for SGX H100 system

sholeksandr:


sholeksandr:
RERUN

PR #414, opened 2022-05-04
Add platform SN3750SX

ciju-nvidia:
This is a new platform, code named Aerial/Komodo. This is based on
Anaconda(MSN3700) platform and the configurations are similar.
It is having secure boot enabled and support for PTP, SyncE, & PPS.

Platform has:
 - NVIDIA Mellanox Spectrum-2 ASIC, 32x200GbE, 6.4Tbps
 - Secure BIOS / CPLD
 - 6 Fans, N+1, hot-swap, forward & reverse
 - 2 PSUs, 1+1 redundancy, hot-swap
 - 16GB DDR4 RAM
 - Intel x86 Coffee Lake E-2254ME CPU

### DIFF
--- a/usr/usr/bin/hw-management.sh
+++ b/usr/usr/bin/hw-management.sh
@@ -188,6 +188,12 @@ msn37xx_secured_connect_table=(    max11603 0x64 5 \
 			tmp102 0x4a 7 \
 			24c512 0x51 8)
 
+sn3750sx_secured_connect_table=(	mp2975 0x62 5 \
+			mp2975 0x66 5 \
+			tmp102 0x49 7 \
+			adt75 0x4a 7 \
+			24c512 0x51 8)
+
 msn3420_base_connect_table=(	max11603 0x6d 5 \
 			xdpe12284 0x62 5 \
 			xdpe12284 0x64 5 \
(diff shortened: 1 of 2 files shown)


usr/usr/bin/hw-management.sh
```diff
@@ -188,6 +188,12 @@ msn37xx_secured_connect_table=(    max11603 0x64 5 \
 			tmp102 0x4a 7 \
 			24c512 0x51 8)
 
+sn3750sx_secured_connect_table=(	mp2975 0x62 5 \
+			mp2975 0x66 5 \
+			tmp102 0x49 7 \
+			adt75 0x4a 7 \
```
sholeksandr, line 194: Are you sure about using 2 different thermal sensor types?

ciju-nvidia:
Yes.. We had a discussion with h/w team, and they confirmed the change of only a single temperature sensor.

Thanks
Ciju
________________________________

sholeksandr (approved):

sholeksandr:
RERUN

usr/usr/bin/hw-management.sh
```diff
@@ -830,6 +836,24 @@ mqmxxx_msn37x_msn34x_specific()
 	get_i2c_bus_frequency_default
 }
 
+sn3750sx_specific()
+{
```
vadimp-nvidia, line 840: Ciju,
Which COME module we connect?
This system should be CFL COME based and original Anaconda was BWL COME based.

ciju-nvidia: It would be CFL_CPU.  It is handled in the function **add_cpu_board_to_connection_table()**. Depending upon the variable $cpu_type, comex module is chosen. As I understand, cpu_type is set in the function **check_cpu_type()** defined in _hw-management-helpers.sh_.

PR #423, opened 2022-05-16
Add BWD COMEX A1 support to msn3700 system.

sholeksandr:
Add BWD COMEX A1 support to msn3700 system.


sholeksandr:
RERUN

PR #439, opened 2022-06-01
hw-mgmt: thermal: Change fan maxumum speed for SN4800

vadimp-nvidia:
Change from 22000 to 18000 RPM.


vadimp-nvidia:
RERUN

PR #441, opened 2022-06-01
hw-mgmt: global wp: Extend global write protection script

sholeksandr:
Add support P4697 system with separate {}_wp bits per ASIC

### DIFF
--- a/usr/usr/bin/hw-management-global-wp.sh
+++ b/usr/usr/bin/hw-management-global-wp.sh
@@ -103,6 +103,53 @@ do_global_wp_release_restore()
 	fi
 }
 
+
+do_asic_wp_release_restore()
+{
+	action="$1"
+	file="$2"
+	command="$3"
+	param1="$4"
+	param2="$5"
+	param3="$6"
+	param4="$7"
+	
+	if [ ! -L "$system_path"/erot1_wp ] || [ ! -L "$system_path"/erot2_wp ]; then
+		return 1
+	fi
+	
+	case "$action" in
+		release)
+			echo 0 > "$system_path"/erot1_wp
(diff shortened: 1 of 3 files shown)


usr/usr/bin/hw-management-global-wp.sh
```diff
@@ -103,6 +103,53 @@ do_global_wp_release_restore()
 	fi
 }
 
+
```
ciju-nvidia, line 106: May be a description would be good here

usr/usr/bin/hw-management-global-wp.sh
```diff
@@ -138,7 +185,13 @@ release|restore)
 		exit 1
 	fi
 	echo "$global_wp_pid" > /var/run/hw-management-global-wp.pid
-	do_global_wp_release_restore "$action" "$2" "$3" "$4" "$5" "$6" "$7"
+	# systems without global WP but with ASIC WP
+	sku=$(< /sys/devices/virtual/dmi/id/product_sku)
```
ciju-nvidia, line 189: Does it make sense to check the if 'sku' is valid?

vadimp-nvidia (approved):

vadimp-nvidia:
RERUN

usr/usr/bin/hw-management-global-wp.sh
sholeksandr: No. 
1. We running do_asic_wp_release_restore onlu if SKU == HI142. In all other cases we running default do_global_wp_release_restore.
2. usr/usr/bin/hw-management-global-wp.sh can by called only on systems with SKU present.

ciju-nvidia (dismissed):

PR #442, opened 2022-06-02
hw-mgmt: scripts: Add dynamic find of gpiobase on SN2201.

mshych:
SN2201 has 2 gpiochips: CPU Intel Denverton pinctrl and PCA9555 I2C GPIO extender.
Both gpiochips have gpiobase that doesn't begin from 0 and more than that this gpiobase
can be different in different kernel versions. E.g., it's different between 5.10.43 and 5.10.103
This commit adds dynamic finding of appropriate gpiobase and use it for further work.

Change also shell scripts access mode to executable.


mshych:
RERUN

vadimp-nvidia (approved):

PR #443, opened 2022-06-02
hw-mgmt: scripts: create in all cases psu...vout attributes

mshych:
Historically there was a case for creating vout2 or vout psu attributes.
Change this for creating always vout attributes.


mshych:
RERUN

vadimp-nvidia (dismissed):

PR #452, opened 2022-06-16
hw-mgmt: scripts: report CPU core temp on SN2201 always for core 0, 1

mshych:
Intel Denverton CPU has strange and not consistent numbering of Cores.
They can be numbered 6, 16 and in other batch of the same architecture they can be 4, 14 etc.
Create hw-management links and sensors rules report always for core 0, 1 as on all other
platforms with different CPU.


mshych:
RERUN

vadimp-nvidia (approved):

mshych:
RERUN

mshych:
RERUN

PR #453, opened 2022-06-16
hw-mgmt: kernel patches: downstream: mlxsw: i2c: Prevent transaction …

vadimp-nvidia:
…execution for special chip states

Do not run transaction in cases chip is in reset or in-service update
states.
In such case firmware is not accessible and will reject transaction
with the relevant status "RUNNING_RESET" or "FW_ISSU_ONGOING".
In case transaction is failed do to one of these reasons, stop sending
transactions. In such case driver is about to be removed since it
cannot continue running after reset or in-service update. And
re-probed again after reset or in-service update is completed.


vadimp-nvidia:
RERUN

mshych:
RERUN

mshych (approved):

PR #454, opened 2022-06-16
hw-mgmt: kernel patches 4.19 & 5.10: align reset_aux_pwr_or_ref attribute name on SN2201

mshych:


mshych:
RERUN

PR #457, opened 2022-06-19
hw-mgmt: kernel patches v4.19: Fixes for line card insertion/removing

vadimp-nvidia:
(1)
platform/mellanox: mlxreg-lc: Fix cleanup on failure and more verbosity in error flow

Clean client object in case of probing failure.
Prevent running remove routine in case probing failed.
Add error log for each kind of failures during probing.

Fixes: 62f9529b8d5c ("platform/mellanox: mlxreg-lc: Add initial support for Nvidia line card devices")

(2)
TMP: DBG: platform/mellanox: mlxreg-lc: Add debug delay
Add debug delay prior the first access to line card.


vadimp-nvidia:
RERUN

felixradensky (approved):

PR #458, opened 2022-06-20
hw-mgmt: topology: Fix A2D i2c sensors tree for mqm8700/msn4600 systems

sholeksandr:


sholeksandr:
RERUN

PR #463, opened 2022-06-22
hw-mgmt: kernel patches v4.19 & v5.10: Align emc2305 driver with mlxreg-fan driver.

mshych:
Set the same manner of setting cooling_cur_state in SN2201 as on other systems,
where it's done through mlxreg_fan driver.


mshych:
RERUN

vadimp-nvidia (approved):

PR #467, opened 2022-06-29
Fixing the incorrect cpld entries

ciju-nvidia:
This patch fixes the incorrect number of cpld entries in
the '/var/run/hw-management/cpld*_*'. [cpld1..4]_* links are
created irrespective of the platform. For example, this
approach has lead to the existance of cpld4_* links, on a
platform with 3 cpld. This was happening due to a race
condition in which clean up of the unwanted symbolic links
executes before the actual entries are created.

Solution is to create the cpld links based on the platform
parameter 'cpld_num', which tells total number of cpld
per platform.

This addresses the bug 3107709

### DIFF
--- a/usr/usr/bin/hw-management-chassis-events.sh
+++ b/usr/usr/bin/hw-management-chassis-events.sh
@@ -601,6 +601,35 @@ function get_i2c_voltmon_prefix()
 	echo "$voltmon_name"
 }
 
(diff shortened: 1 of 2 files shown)


ciju-nvidia:
RERUN

yyehudai (approved):

felixradensky (approved):

PR #484, opened 2022-07-08
hw-mgmt: topology: Remove A2D i2c sensors from MSN2700/MSN2410 systems Remove A2D i2c sensors from MSN2700/MSN2410 systems management + CPU board

sholeksandr:


sholeksandr:
RERUN

PR #485, opened 2022-07-10
hw-mgmt: patches: v5.10: Skip attempt to access line card INI

vadimp-nvidia:
In upstream code mlxsw driver has ability to burn INI.
This is not relevant for case when mlxsw_minimal driver is co-existing with
Nvidia SDK software, where this file is not supposed to be present.

Disable any attempt to find INI file to avoid output of warning message.


vadimp-nvidia:
RERUN

felixradensky (approved):

PR #492, opened 2022-07-11
hw-mgmt: kernel patches 4.19, 5.10: Reformat downstream leds-mlxreg patch and rename them.

mshych:
Reformat leds-mlgreg patches, so they can be applied with git am.
Rename them to show that these are just downstream patches.


mshych:
RERUN

sholeksandr (approved):

PR #495, opened 2022-07-13
Fix voltmon sensors linking by their labels for mp2975.

sholeksandr:
Fix voltmon sensors linking issue hen we have >= sensors on voltmon


sholeksandr:
RERUN

sholeksandr:
RERUN

PR #498, opened 2022-07-14
hw-mgmt: patches: 5.10: Add OSFP support

vadimp-nvidia:
Add support for OSFP transceiver modules by adding their SFF-8024
Identifier Value (0x19).

This is required for future NVIDIA Spectrum-4 based systems that will be
equipped with OSFP transceivers.


vadimp-nvidia:
RERUN

mshych (approved):

PR #499, opened 2022-07-14
hw-mgmt: topology: Align i2c sensors tree for msn3700 respin A1

sholeksandr:


sholeksandr:
RERUN

PR #501, opened 2022-07-17
hw-mgmt: rules: Add support for netdev renaming on multi-asic systems

felixradensky:
On mutli-asic systems like MQM9510 and MQM9520 the current netdev
renaming udev rule will create conflicts, as each port name appears
on multiple switches. Fix this by incorporating switch name into
port name. Also, move the rule to the end of list to ensure it is
executed last and not overriden by the old MAC-based netdev rules.

### DIFF
--- a/usr/usr/bin/hw-management-if-rename.sh
+++ b/usr/usr/bin/hw-management-if-rename.sh
@@ -1,0 +1,69 @@
+#!/bin/bash
+
+# Copyright (c) 2018 - 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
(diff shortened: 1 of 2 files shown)


mshych (dismissed):

usr/usr/bin/hw-management-if-rename.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
```
vadimp-nvidia, line 2: Need license.

vadimp-nvidia (dismissed):

usr/usr/bin/hw-management-if-rename.sh
felixradensky: Added license

felixradensky:
RERUN

vadimp-nvidia (approved):

PR #509, opened 2022-07-20
hw-mgmt: patches: 4.19 & 5.10: report 0 RPM speed in EMC2305 driver for SN2201 system.

mshych:
1. Support of report 0 RPM in case that tacho read below minimal range.
2. Change Fan minimal to 2200 RPM. This is a real RPM speed when FAN still rotates.


mshych:
RERUN

recipes-kernel/linux/linux-5.10/0087-hwmon-Add-support-for-EMC2305-RPM-based-PWM-Fan-Spee.patch
```diff
@@ -232,7 +234,11 @@ index 000000000..3f98bdcd2
 +		return ret;
 +
 +	ret = ret >> EMC2305_TACH_REGS_UNUSE_BITS;
-+	return EMC2305_RPM_FACTOR * EMC2305_TACH_CNT_MULTIPLIER / (ret > 0 ? ret : 1);
++	ret = EMC2305_RPM_FACTOR * EMC2305_TACH_CNT_MULTIPLIER / (ret > 0 ? ret : 1);
++	if (ret <= EMC2305_TACH_RANGE_MIN * EMC2305_TACH_CNT_MULTIPLIER)
++		return 0;
++	else
++		return ret;
```
vadimp-nvidia, line 241: No need else.
But I think i have one redundant "* EMC2305_TACH_CNT_MULTIPLIER"
Why not to have
if (ret <= EMC2305_TACH_RANGE_MIN)
return 0;
and after that old code
return EMC2305_RPM_FACTOR * EMC2305_TACH_CNT_MULTIPLIER / (ret > 0 ? ret : 1);

mshych: Because read val from reg is 0xfff0. I don't want to use some magic values 0x1ffe but use documented in DS tacho min range = 480. It can be checked just after calculation.

vadimp-nvidia: But this is exactly
if (ret <= EMC2305_TACH_RANGE_MIN)
return 0;

Otherwise you have 
EMC2305_RPM_FACTOR * EMC2305_TACH_CNT_MULTIPLIER
EMC2305_TACH_RANGE_MIN * EMC2305_TACH_CNT_MULTIPLIER
And redundant else

PR #529, opened 2022-07-26
hw-mgmt: attributes: Change temperature sensor model for MQM9510 and …

felixradensky:
…MQM9520

MQM9510 and MQM9520 will be using ADT75 temperature sensors.


felixradensky:
RERUN

mshych (approved):

PR #532, opened 2022-07-28
hw-mgmt: init: Set config parameters before starting udev

felixradensky:
Make sure the configuration parameters are set before udev is
started, otherwise they may be undefined in scripts started by
udev rules.


felixradensky:
RERUN

felixradensky:
RERUN

ciju-nvidia (approved):

PR #536, opened 2022-08-04
hw-mgmt: scripts & conf: introduce SMBIOS alternative BOM mechanism of system description.

mshych:
This new mechanism allows to initialize system with appropriate components at required
location according to received SMBIOS System Version string.
The string is generated from Agile according to actual BOM that is used
for production particular system and it's burned in customization step.
This allows dynamically differentiate same system type but with different
alternative components.
Codes for this SMBIOS string can be found in:
https://nvidia.sharepoint.com/:x:/r/sites/NBU-sw/Systems/Low_Level/_layouts/15/doc2.aspx?sourcedoc=%7B4a10e76f-3c80-48e7-a7f2-4b8bfff83319%7D

This mechanism contains also simualtion and debug tool: hw-management-devtree-check.sh


mshych:
RERUN

PR #540, opened 2022-08-09
hw-mgmt: scripts: changes in links of NVMe SSD temperatures.

mshych:
1. Don't create link to temp1_alarm. This attribute unneeded and undocumented.
2. Check if temp1_min attribute exist before creation link.
Old driver versions of nvme hwmon don't create this attribute.

### DIFF
--- a/usr/usr/bin/hw-management-thermal-events.sh
+++ b/usr/usr/bin/hw-management-thermal-events.sh
@@ -830,10 +830,11 @@ if [ "$1" == "add" ]; then
 					# Make links only to 1st sensor - Composite temperature.
 					# Normaslized composite temperature values are taken to thermal management.
 					if [ "$i" -eq 1 ]; then
-						ln -sf "$3""$4"/temp"$i"_alarm "$thermal_path"/"$dev_name"_"$name"_alarm
 						ln -sf "$3""$4"/temp"$i"_crit "$thermal_path"/"$dev_name"_"$name"_crit
 						ln -sf "$3""$4"/temp"$i"_max "$thermal_path"/"$dev_name"_"$name"_max
-						ln -sf "$3""$4"/temp"$i"_min "$thermal_path"/"$dev_name"_"$name"_min
+						if [ -e "$thermal_path"/"$dev_name"_"$name"_min ]; then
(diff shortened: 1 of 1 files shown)


mshych:
RERUN

ciju-nvidia (approved):
It would create 'nvme_composite_min' in 5.10, but will not create in 4.19. It should be fine if we are not claiming support for Alligator with 4.19 kernel. If we need to claim 4.19 support, better not to create this attribute as we would want to keep the links consistent across 4.19 & 5.10 kernels.

mshych:
No one uses this argument.
This is the max that we can do, not create a broken link.
This missing argument is related to an old driver version. It's happened also on other SSD types.
At all, nvme temp arguments are just nice to have.
We don't have something similar on SATA SSDs, even on SSD that have temp sensor.

felixradensky (approved):

vadimp-nvidia (approved):

usr/usr/bin/hw-management-thermal-events.sh
```diff
@@ -830,10 +830,11 @@ if [ "$1" == "add" ]; then
 					# Make links only to 1st sensor - Composite temperature.
 					# Normaslized composite temperature values are taken to thermal management.
 					if [ "$i" -eq 1 ]; then
-						ln -sf "$3""$4"/temp"$i"_alarm "$thermal_path"/"$dev_name"_"$name"_alarm
 						ln -sf "$3""$4"/temp"$i"_crit "$thermal_path"/"$dev_name"_"$name"_crit
 						ln -sf "$3""$4"/temp"$i"_max "$thermal_path"/"$dev_name"_"$name"_max
-						ln -sf "$3""$4"/temp"$i"_min "$thermal_path"/"$dev_name"_"$name"_min
+						if [ -e "$thermal_path"/"$dev_name"_"$name"_min ]; then
```
sholeksandr, line 835: Not understand this condition, You for existilg links that you want to creste.
I think it should be
if [ -e  "$3""$4"/temp"$i"_min  ]; then
Ir better to use check_n_link from helpers
check_n_link  "$3""$4"/temp"$i"_min "$thermal_path"/"$dev_name"_"$name"_min
check_n_link

mshych: Oh, you are right. Thx

PR #551, opened 2022-08-23
hw-mgmt: infra: Fix PCI to I2C ASIC bus mapping on multi-asic systems

felixradensky:

### DIFF
--- a/usr/usr/bin/hw-management.sh
+++ b/usr/usr/bin/hw-management.sh
@@ -2153,20 +2153,10 @@ do_chip_up_down()
 
 	map_asic_pci_to_i2c_bus $pci_bus
 	bus=$?
-	# TMP: Use this mapping for getting bus below. Modify it as following.
-	#if [ $bus -eq 0 ]; the
-	#	# Add ASIC device.
-	#	if [ asic_index -eq 1 ]; then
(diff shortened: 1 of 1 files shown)


felixradensky:
RERUN

usr/usr/bin/hw-management.sh
```diff
@@ -2153,20 +2153,10 @@ do_chip_up_down()
...
-	#	else
-	#		get_asic2_bus
-	#	fi
-	#	bus=$?
-	#fi
-	# Remove three line below after TMP is uncommented.
-	# Add ASIC device.
-	get_asic_bus
-	bus=$?
+	if [ $bus -eq 0 ]; then
```
vadimp-nvidia, line 2156: I think we need validate we are OK with Gorilla or Leopard in OPT, when we make sdk_start.sh.
We should be OK, but better to check we don't break. In our regression in ONL we don't run sdk_start.sh/sdk_stop.sh.
So we can miss something.

vadimp-nvidia (approved):

usr/usr/bin/hw-management.sh
felixradensky: OK, will check

PR #554, opened 2022-08-31
hw-mgmt: rules: Rename line card interfaces

vadimp-nvidia:
Remove line card index from name


sholeksandr:
RERUN

sholeksandr (approved):

PR #563, opened 2022-09-02
Fixing the psu attributes on hw-management restart

ciju-nvidia:
This patch fixes the missed out config/psu[1-2]_eeprom_us attributes
after the hw-management service is restarted. This issue happens
because psu controllers are not removed during the hw-management stop.
The corresponding eeprom device files
/sys/devices/platform/mlxplat/i2c_mlxcpld.1/i2c-1/i2c-4/4-00[50-51]/eeprom
are also not removed.

Bugs Fixed: 3168642


vadimp-nvidia (approved):

ciju-nvidia:
RERUN

PR #573, opened 2022-09-13
hw-mgmt: scripts: fix CFL commex bus in SMBios/devtree og legacy systems.

mshych:


mshych:
RERUN

PR #578, opened 2022-09-21
hw-mgmt: init: Add retry mechanism for regio detection

vadimp-nvidia:
In some marginal cases during early initialization sysfs entry name for mlxreg-io driver can be created with some small delay. In such case the below error is output to syslog:
"ERR hw-management: mlxreg_io is not loaded".

Add retry mechanism to protect from such timing issue.

Fixes: [SONIC - Design - Bug SW #3177482] [Non-Functional ] [HW-mgmt] [error in log] | ERR hw-management: mlxreg_io is not loaded

### DIFF
--- a/usr/usr/bin/hw-management.sh
+++ b/usr/usr/bin/hw-management.sh
@@ -508,7 +508,7 @@ function restore_i2c_bus_frequency_default()
 	esac
 }
 
-function find_regio_sysfs_path()
+function find_regio_sysfs_path_helper()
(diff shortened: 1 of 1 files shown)


sholeksandr (approved):

vadimp-nvidia:
RERUN

ciju-nvidia (approved):

vadimp-nvidia:
RERUN

PR #584, opened 2022-09-29
hw-mgmt: thermal: Fix unexpected high FAN RPM value during TC start

sholeksandr:
When TC starts, it checks the state of thermal zones. If some TZ
returns a wrong (zero) value in temp_trip_high together with the valid
temperature, then TC not returning the cooling level to normal value.
This fix adds stage of "re-init" cooling level value after temp_trip_high
value becomes normal

Fixes: RM #3200958 : Unexpected high FAN RPM value with CL5.2


sholeksandr:
RERUN

sholeksandr:
RERUN

yyehudai (approved):

PR #591, opened 2022-10-25
hw-mgmt: init: Complete init flow only after the last attribute is cr…

vadimp-nvidia:
…eated

Wait in initiailization flwo until  'cpld_base' is created internally after all other links are created. It combines few attributes together, so it is created at the end.

Fixes: [No RM] ("degradation in hw-mgmt V.7.0020.4001 on 2700: it takes more time to link the reboot cause files")

### DIFF
--- a/usr/usr/bin/hw-management-start-post.sh
+++ b/usr/usr/bin/hw-management-start-post.sh
@@ -54,3 +54,7 @@ case $board in
 		;;
 esac
 
+if [ ! -f /var/run/hw-management/system/cpld_base ]; then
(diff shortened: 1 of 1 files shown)


vadimp-nvidia:
RERUN

mshych:
What do you mean by "...eated" in a commit?
Maybe decrease TO to 30sec from 300?

mshych (approved):

vadimp-nvidia:
RERUN

vadimp-nvidia:
RERUN

PR #594, opened 2022-10-26
hw-mgmt: sensors.conf: changes of SN5600 sensors configuration.

mshych:
1. Ignore exception VLs (in2) of Delta power-converter.
   This input isn't reported in standard linear form by vendor.
   It still can be read by raw pmbus command, not through pmbus driver, and converted
   by exception formula provided by Delta.
2. Beautify: fix indentations.


mshych:
RERUN

vadimp-nvidia (approved):

PR #598, opened 2022-11-02
hw-mgmt: psu: Add routine to set PSU fan speed

felixradensky:
For most PSUs we assume that fan speed is always given in duty cycle (percentage) units. This may not always be the case, for example the PSU FW default units may be RPM, or the units can be modified by kernel driver or userspace tools. Create a routine to set PSU fan speed that expicitly sets speed units and use it at all places where PSU fan speed is being set.


felixradensky:
RERUN

vadimp-nvidia (approved):

PR #602, opened 2022-11-07
hw-mgmt: topology: Fix voltmon i2c bus addresses for P4697 system Fix mp2975 voltmon's i2c bus addresses for P4697 system

sholeksandr:


sholeksandr:
RERUN

PR #603, opened 2022-11-08
hw-mgmt: psu: Add special FW update handling for first batch of Delta…

felixradensky:
… 3K PSUs

The first batch of Delta 3K PSUs uses a different PMBus command to check the FW upload status. Add special handling for this PSU model.


felixradensky:
RERUN

mshych (approved):

PR #604, opened 2022-11-08
hw-mgmt init & docs: remove icp201xx pressure sensor from common hw-mgmt handling

mshych:


mshych:
RERUN

ciju-nvidia (approved):

PR #608, opened 2022-11-13
hw-mgmt: attributes: Avoid reading cpld3 version om MSB78xx systems

felixradensky:
There are only 2 CPLDs on MSB78xx systems, do not create cpld_port attribute to avoid reading cpld3 version on these systems. In case reading cpld3 version is attempted anyway, (e.g. due to error in system identification), do not display error message.


felixradensky:
RERUN

felixradensky:
RERUN

felixradensky:
RERUN

vadimp-nvidia (approved):

PR #612, opened 2022-11-17
hw-mgmt: scripts: add settings for power calculation on SN5600

mshych:

### DIFF
--- a/usr/usr/bin/hw-management-thermal-events.sh
+++ b/usr/usr/bin/hw-management-thermal-events.sh
@@ -812,6 +812,16 @@ if [ "$1" == "add" ]; then
 			fw_ver="N/A"
 			if [ "$cap" == "550" -o "$cap" == "2000" -o "$cap" == "3000" ]; then
 				fw_ver=$(hw_management_psu_fw_update_delta.py -v -b $bus -a $psu_addr)
+				if [ "$cap" == "3000" ] && [ "$board_type" == "VMOD0013" ]; then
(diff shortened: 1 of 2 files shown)


mshych:
RERUN

usr/usr/bin/hw-management-thermal-events.sh
```diff
@@ -1156,6 +1166,11 @@ else
 		psu_disconnect_power_sensor "$psu_name"_curr
 
 		rm -f $eeprom_path/"$psu_name"_vpd
+		rm -f $fw_path/"$psu_name"_fw_ver
+		if [ -e "$config_path"/"$psu_name"_power_slope ]; then
+			rm -f "$config_path"/"$psu_name"_power_slope
+			rm -f "$config_path"/"$psu_name"_power_capacity
```
ciju-nvidia, line 1172: Just checking if there is a need to remove   "$config_path"/amb_tmp_crit_limit   &  $config_path"/amb_tmp_crit_limit

mshych: Yes, they are created together so there is no case if there is power_slope and no power_capacity

mshych:
RERUN

usr/usr/bin/hw-management-thermal-events.sh
mshych: An no need to remove "$config_path"/amb_tmp_crit_limit & $config_path"/amb_tmp_crit_limit as they are common for both PSUs.

ciju-nvidia (approved):

PR #616, opened 2022-11-21
hw-mgmt: events: Run chipup flow triggered by ASIC health event only …

vadimp-nvidia:
…in specific OS

There are two automatic chipup/chipdown flows:
Based on SDK trigger.
Based on ASIC health event.
Run the second one only in ONL OS. In other OSes it will be triggered by SDK.

Use match for onl folder instead of match of SDK folder. Remove redundant logic for unknown health state.

Fixes: #3263106: Minimal driver probing of 2-0048 failed with error -110: During boot when no SDK Detected


vadimp-nvidia:
RERUN

sholeksandr (approved):

vadimp-nvidia:
RERUN

sholeksandr:
RERUN

PR #623, opened 2022-11-28
hw-mgmt: events: Modify chipup flow triggered by ASIC health event on…

vadimp-nvidia:
…ly in specific OS

Make this flow OS independent.

Use special flag for activation, when required. This flag is provided by OS.


vadimp-nvidia:
RERUN

sholeksandr (approved):

PR #631, opened 2022-12-15
hw-mgmt: psu: Add psuX_fan_dir attribute

sholeksandr:
Added psuX_fan_dir attrbute to {hw-management}/thermal/ folder
Value in this file represents PSU fan direction:

 0 - Forward (C2P)
 1 - Reverse (P2C)
 2 - unknown (read error or field missing)

PSU FAN direction value can be extracted from PSU VPD, which can be in 2 formats:
 1. PN_VPD_FIELD: MTEF-PSF-AC-G
 2. PN_VPD_FIELD: 930-9SPSU-00RA-00B

### DIFF
--- a/usr/usr/bin/hw-management-thermal-events.sh
+++ b/usr/usr/bin/hw-management-thermal-events.sh
@@ -200,6 +200,45 @@ get_fixed_fans_direction()
 	esac
 }
 
+# Get PSU direction based on VPD PN field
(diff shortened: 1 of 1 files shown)


yyehudai (approved):

mshych (commented):
I think that create fan direction file with error code is misleading.
Maybe better not to create this file at all and just report parsing error in logs

PR #643, opened 2022-12-20
hw-mgmt: patches 5.10: Add patches blacklist for 5.10.103

sholeksandr:
Added blacklist for kernel patches which should not be appyied on kernel 5.10.103 (Sonic build)


sholeksandr:
RERUN

PR #646, opened 2022-12-21
hw-mgmt: thermal: Increase fan minimum speed after suspend

vadimp-nvidia:
Set fan not below 40% after thermal control is resumed.


vadimp-nvidia:
RERUN

mshych (approved):

PR #649, opened 2022-12-28
hw-mgmt: thermal: Proivide step-by-step fan speed download during boo…

vadimp-nvidia:
…t at restart

During initialization decrease fan speed stepping down 10% from full speed down to dynamnic minimum.


vadimp-nvidia:
RERUN

PR #657, opened 2023-01-09
hw-mgmt: kernel patches: Add patch for extending hotplug confuguration

vadimp-nvidia:
Currently hotplug configuration in logic device assumes that all items are provided with no holes.
Thus, any group of hotplug events, associated with the specific status/event/mask registers is configured in those registers successively from bit zero to bit #n (#n < 8).

This logic is changed int order to allow non-successive definition to support configuration with the skipped bits – for example bits 3, 5, 7 in status/event/mask registers can be associated with hotplug events, while others can be skipped.


vadimp-nvidia:
RERUN

mshych (dismissed):

vadimp-nvidia:
RERUN

vadimp-nvidia:
RERUN

vadimp-nvidia:
RERUN

vadimp-nvidia:
RERUN

mshych (approved):

PR #661, opened 2023-01-17
hw-mgmt: fan: Fix fan dir attribute for systems without CPLD fan direction attribute

sholeksandr:


sholeksandr:
RERUN

vadimp-nvidia (approved):

PR #662, opened 2023-01-18
hw-mgmt: psu: Fix reporting of Delta 550 FW version

felixradensky:
Delta 550 FW version reporting format (XX.YY.ZZ) differs from other Delta PSUs (2000/3000), which report FW version as XX.YY. Add check to detect 550 model and report its FW version correctly.


vadimp-nvidia (approved):

felixradensky:
RERUN

PR #681, opened 2023-02-01
Capturing cpld dump in the hw-management dump

ciju-nvidia:
This patch adds the support for cpld dump capture in the hw-management dump generation script.


ciju-nvidia:
RERUN

felixradensky (commented):
Hi Ciju,
This change will not work on ARM where iorw is not available, you should use iorw.sh instead

ciju-nvidia:
RERUN

ciju-nvidia:
RERUN

felixradensky (approved):

PR #686, opened 2023-02-05
hw-mgmt: kernel patches: Modify pressure sensor patch

vadimp-nvidia:
Modify downstream patch #9000 (pressure sensor)


vadimp-nvidia:
RERUN

mshych (approved):

PR #688, opened 2023-02-06
hw-mgmt: scripts: comment out old systems from SMBIOS BOM support.

mshych:


mshych:
RERUN

vadimp-nvidia (approved):

PR #697, opened 2023-02-20
hw-magmt: scripts: Fix FAN dir attribute for systems with fixed FANs

sholeksandr:
On some old systems (2100, 2010) with fixed FANs we are getting FAN
direction based on MLNX PN in the main board FRU. If this FRU is
broken or PN was changed we can't get FAN direction.This commit adds an
additional type for FAN dir detection - by the ambient sensors themperature
comparison.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,4 @@
-hw-management (1.mlnx.7.0020.5135) unstable; urgency=low
+hw-management (1.mlnx.7.0020.5038) unstable; urgency=low
   [ MLNX ] 
 
- -- MellanoxBSP <system-sw-low-level@mellanox.com>  Sun, 1 Jan 2022 11:55:00 +0300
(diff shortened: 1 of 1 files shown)


usr/usr/bin/hw-management-thermal-events.sh
```diff
@@ -219,7 +219,13 @@ get_psu_fan_direction()
 	dir_char=""
 	pn="$(grep PN_VPD_FIELD $vpd_file)"
 	if [ -z $pn ]; then
-		return 2
+		if [ -f $config_path/fixed_fans_dir ]; then
+			dir=$(< $config_path/fixed_fans_dir) 
+		else
+			# Default dir P2C till it will not be detected later
```
mshych, line 225: It's not P2C according to your code and following comment.
Pls change this comment.

PR #699, opened 2023-02-21
hw-mgmt: Fix detection of BlueField-3 CPU type

felixradensky:
BlueField-3 uses Cortex-A78AE ARM cores, which have 0xD42 P/N.


mshych (approved):

felixradensky:
RERUN

PR #709, opened 2023-02-27
hw-mgmt: sensors: Add WA for unsupported PSU command in SN5600 config

felixradensky:
Delta 3K PSU does not support 0x31 (POUT_MAX) PMBus command. In response to this command it returns 0xff 0xff as data and 0xff as PEC. For PSU on address 0x5a the calculated PEC is indeed 0xff, which causes linux PMBus PSU driver to recognize this command as supported and create power2_cap sysfs attribute with some invalid data.

Add WA to show the value of this attribute as 0 in the output of "sensors" command.


felixradensky:
RERUN

mshych (approved):

PR #739, opened 2023-03-16
hw-mgmt: scripts: LED color attribute change

ciju-nvidia:
In newer switches, the LED color is amber and not orange. This patch renames the LED attributes that contains the string 'orange' to 'amber'. Also updates the capability string from 'orange' to 'amber'.


ciju-nvidia:
RERUN

felixradensky (approved):

PR #744, opened 2023-03-21
hw-mgmt: thermal: Initial commit for thermal_control 2.0

sholeksandr:
This commit introduces a new version of thermal control which is fully working
in the user space.

This new TC is based on python3 engine and and running in user space.
It can be controlled as a service:
    service hw-management-tc start/stop/status


sholeksandr:
RERUN

sholeksandr:
RERUN

yyehudai (approved):

PR #745, opened 2023-03-21
hw-mgmt: kernel patches: Fix deploy patch tool

sholeksandr:
1. Added sonic specific patch 0052
2. Set command line option --config to optional


sholeksandr:
RERUN

PR #746, opened 2023-03-23
hw-mgmt: thermal: Add TC service start by default

sholeksandr:


sholeksandr:
RERUN

PR #749, opened 2023-03-27
hw-mgmt: kernel patches: Fix kernel config options

sholeksandr:
Fix kernel config options file which is used by the patch deploy tool

CONFIG_I2C=m            =>  CONFIG_I2C=y
CONFIG_PINCTRL_INTEL=m  =>  CONFIG_PINCTRL_INTEL=y


sholeksandr:
RERUN

PR #755, opened 2023-03-29
hw-mgmt: sensors: fix swapped addresses of PSUs in SN5600

mshych:


mshych:
RERUN

PR #760, opened 2023-04-04
hw-mgmt: fan: Fix tacho order for inversed FAN order

sholeksandr:
Fix tacho order for inversed FAN order
bug: 3430203


sholeksandr:
RERUN

yyehudai (approved):

PR #761, opened 2023-04-04
hw-mgmt: patches 5.10: Align patches with upstream

sholeksandr:
1. Align patches with upstream
2. Add TMP patch for FAN speed fix


sholeksandr:
RERUN

PR #766, opened 2023-04-20
hw-mgmt: scripts: add new script for check results of latest performed BIOS update.

mshych:


mshych:
RERUN

mshych:
RERUN

PR #768, opened 2023-04-24
hw-mgmt: thermal: Change dmin values for 2100/2010 systems

sholeksandr:
Change dmin table values for 2100/2010. Set it to same values for P2C/C2P dir.
It will fix TC algo in case we can't read FAN dir.


sholeksandr:
RERUN

ciju-nvidia (approved):

yyehudai (approved):

PR #774, opened 2023-04-28
hw-mgmt: thermal: Fix bug "Network is unreachable" in logger

sholeksandr:
Fixing bug in syslog logging "Network is unreachable" when
the ip address on the lo interface gets deleted

Bug: #3445789


sholeksandr:
RERUN

yyehudai (approved):

PR #782, opened 2023-05-09
hw-mgmt: kernel 5.10: add version, PN and minor version registers of 5th CPLD.

mshych:
Add report about 5th CPLD. SN5400 will have 5 CPLDs.


mshych:
RERUN

vadimp-nvidia (approved):

PR #783, opened 2023-05-11
hw-mgmt: Remove ICP201xx pressure sensors from SMBIOS BOM mechanism.

mshych:
These pressure sensors don't have upstream kernel driver.
They will be instantiated manually in OPT-OS only.


mshych:
RERUN

sholeksandr (approved):

mshych:
RERUN

PR #790, opened 2023-05-23
hw-mgmt: thermal: Update default config for PWM/FAN control

sholeksandr:
Set PWM/FAN config to not use ASIC control by default


sholeksandr:
RERUN

PR #791, opened 2023-05-25
hw-mgmt: thermal: Fix TC log error on service stop

sholeksandr:


sholeksandr:
RERUN

PR #798, opened 2023-06-02
hw-mgmt: topology: Fix MQM9700 sensors.conf with xdpe12284

sholeksandr:
Fix MQM9700 sensors.conf with xdpe12284 voltmon


sholeksandr:
RERUN

PR #802, opened 2023-06-08
V.7.0030.1000 BR

sholeksandr:
- hw-mgmt: debug: Fix copy file filter for hw-management scripts
- hw-mgmt: thermal: fixes for TC


sholeksandr:
RERUN

PR #806, opened 2023-06-17
hw-mgmt: kernel patches: Align patches to support 5.10.179

sholeksandr:
Align patches to support 5.10.179


sholeksandr:
RERUN

PR #807, opened 2023-06-20
hw-mgmt: patches 5.10: Fix for asic sensor count

ciju-nvidia:
In newer platforms like Komodo there is an additional OCXO temperature sensor. This sensor is enabled in the firmware and the platform driver returns the SFP module number which starts with three and causes the
module*_temperature attributes to start with module2. This patch fixes the numbering issue.


ciju-nvidia:
RERUN

vadimp-nvidia (approved):

felixradensky (commented):
1. The example in patch description seems incorrect
2. There's a typo in comment in hw-management-sensor-events.sh (sensros instead of sensors)

ciju-nvidia:
RERUN

PR #811, opened 2023-06-25
master

sholeksandr:
- hw-mgmt: Add support for mlx-platform driver on BF3 systems
- hw-mgmt: init: Fix hw-management init


sholeksandr:
RERUN

PR #814, opened 2023-06-27
hw-mgmt: thermal: Fix cpu_pack "sensor reading error"

sholeksandr:
Fix cpu_pack sensor reading error handle.


sholeksandr:
RERUN

PR #821, opened 2023-07-05
hw-mgmt: Add indicator for VPD parsing completion

vadimp-nvidia:
Set attributes "events_ready" to 1 after VPD parsing is done.


vadimp-nvidia:
RERUN

felixradensky (approved):

PR #824, opened 2023-07-06
Fix for Voltmon temperature attribute in Gorilla

ciju-nvidia:
The attributes voltmon*_temp_input is not created and this patch fixes that.

Issue# 3530045


ciju-nvidia:
RERUN

vadimp-nvidia (approved):

PR #825, opened 2023-07-06
 Loading modules based on the availability 

ciju-nvidia:
Some drivers are not available for older kernels. For
example drivetemp module is not present in 4.19 kernel.
This patch adds a function to load such modules based
on their availability in a specific kernel version.

Issue: #3516754

### DIFF
--- a/usr/usr/bin/hw-management.sh
+++ b/usr/usr/bin/hw-management.sh
@@ -2065,6 +2065,19 @@ get_asic2_bus()
 	return $((asic_bus))
 }
 
+load_modules()
+{
+	depmod -a 2>/dev/null
+	# Some modules are not present in all the kernel
+	# versions. Use this function to load those modules
+	# which need to be loaded based on their availability
+	if ! lsmod | grep -q "drivetemp"; then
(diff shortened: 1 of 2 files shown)


ciju-nvidia:
RERUN

felixradensky (approved):

ciju-nvidia:
RERUN

usr/usr/bin/hw-management.sh
```diff
@@ -2328,6 +2340,7 @@ do_start()
 		get_asic2_bus
 	fi
 	touch $udev_ready
+	load_modules
```
mshych, line 2343: It's too close to udev retrigger where event from this driver should be handled.
Also, it looks more logical to make this before udev_ready.

mshych (commented):
It's too close to udev retrigger where event from this driver should be handled.
Also, it looks more logical to make this before udev_ready.

usr/usr/bin/hw-management.sh
ciju-nvidia: Please see the updated version.

ciju-nvidia:
RERUN

ciju-nvidia:
RERUN

ciju-nvidia:
RERUN

PR #827, opened 2023-07-09
hw-mgmt: Add resiliency mechanism for chipup operation

vadimp-nvidia:
Validate after 'chipup' that 'hwmon' object is created. If it is not - retry operation by 'chipdown' / 'chipup' up to 3 times.

For manual testing runat the same time:
hw-management.sh chipup &
/etc/init.d/sxdkernel start &

The second operation should reset ASIC and causing failure of the 1-st operation.
Retry mechanism should recover this failure.


vadimp-nvidia:
RERUN

yyehudai (approved):

PR #833, opened 2023-07-11
Fix performance issue in less powerful cpu systems

ciju-nvidia:
Label maker script is invoked while creating the symbolic links during the udev events. In old systems which are having less powerful cpu, it's creating some performance
overhead. This patch moves the sku check to one levl above so that label maker script process will not be invoked.

### DIFF
--- a/usr/usr/bin/hw-management-helpers.sh
+++ b/usr/usr/bin/hw-management-helpers.sh
@@ -168,6 +168,15 @@ unlock_service_state_change()
     /usr/bin/flock -u ${LOCKFD}
 }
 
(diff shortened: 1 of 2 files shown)


vadimp-nvidia (approved):

felixradensky (commented):
check_labels_enabled() should return 0 for gorilla and 1 for all others. In shell scripts 0 means success. Then you can change 
if [ $(check_labels_enabled) -eq 1 ]; then
by
if check_labels_enabled; then

sholeksandr (approved):

PR #834, opened 2023-07-13
hw-mgmt: patches 5.10: Add ARCH select support to patch_deploy_tool

sholeksandr:
We have different kernel CONFIG_* sets for different ARCH.
With this commit, we added the option to select ARCH in patch_deploy_tool

--arch {amd64,arm64}  Arch type amd64/arm64/... (default: amd64)


sholeksandr:
RERUN

felixradensky (commented):
CONFIG_GPIO_MLXBF3=m should be removed for ARM64 because relevant patch is currently missing. Also please remove the following options, they are not relevant for BF3, only for BF1 and BF2:
CONFIG_MLXBF_LIVEFISH=m
CONFIG_MLXBF_TRIO=m

PR #838, opened 2023-07-17
hw-mgmt: patches kernel 5.10: fix timing issue of add FAN EEPROMs on SN2201.

mshych:
Define the exact i2c bus (adapter number) of fans on the SN2201 system.
This will cause fan's EEPROMs be connected already from nvsw-sn2201
platform driver and not from user space after receiving udev events.


mshych:
RERUN

vadimp-nvidia (approved):

felixradensky (approved):

sholeksandr (approved):

PR #846, opened 2023-07-20
hw-mgmt: thermal: fix TC start for unsupported systems

sholeksandr:
Fix TC service error for unsupported systems

Bug: #3517930
"fanless system fails hw-management-tc.service at boot up and also fails to restart after boot up"


sholeksandr:
RERUN

ciju-nvidia (approved):

PR #848, opened 2023-07-20
V.7.0030.1000 BR

sholeksandr:
- hw-mgmt: thermal: add reload TC service on crash
- hw-mgmt: thermal: fix TC start for unsupported systems


sholeksandr:
RERUN

ciju-nvidia (approved):

PR #853, opened 2023-07-26
hw-mgmt: scripts: Fix unnecessary PSUs volt_out2 links

mshych:
BUG# 3549280

Don't create automatically volt_out2 link for PSU in3 attribute
but just in case that it's label ssn't vout1.


mshych:
RERUN

felixradensky (approved):

PR #856, opened 2023-07-31
hw-mgmt: thermal: Fix msn6560 tc config data link

sholeksandr:
Bug: #3559366


sholeksandr:
RERUN

PR #865, opened 2023-08-02
hw-mgmt: thermal: Fix ASIC tz mode

sholeksandr:
In case if ASIC starting after TC it can set thermal zone mode to
'step_wise'.
This fix set thermal zone polict to 'user_space' after ASIC start.


sholeksandr:
RERUN

sholeksandr:
RERUN

felixradensky (approved):
In commit message s/

PR #866, opened 2023-08-02
hw-mgmt: patches: Initial commit for v6.1

vadimp-nvidia:
Add patches to support kerenl v6.1 (rebased on top of v6.1.42).


sholeksandr:
RERUN

PR #870, opened 2023-08-03
hw-mgmt: thermal: Disable kernel thermal algorithm

sholeksandr:
Disable kernel thermal algorithm for all systems


sholeksandr:
RERUN

PR #885, opened 2023-08-09
hw-mgmt: thermal: Add static sensor support

sholeksandr:
On init flow, TC scans available sensors and adds them to DB. Some sensors can be present or not depending on system types. For example sodimm2_temp present not in all systems, and sensors scan will add it only if exists sensor link in sysfs. But if the system should have sodimm2 but it is broken, we will not recognize such an issue.

Static sensor - a sensor that should be added, even if HW is not present or broken. Static sensor list can be added to TC system data JSON file:

"static_sensor_list": ["sodimm2", "sensor_amb", "cpu_pack"]

Bug: #3569409

### DIFF
--- a/usr/etc/hw-management-thermal/tc_config_default.json
+++ b/usr/etc/hw-management-thermal/tc_config_default.json
@@ -1,4 +1,5 @@
  {
+	"name": "default",
 	"dmin" : {
 		"C2P": {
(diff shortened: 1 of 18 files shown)


yyehudai (approved):

sholeksandr:
RERUN

sholeksandr:
RERUN

PR #900, opened 2023-08-15
hw-mgmt: scripts: add new reset ready attribute.

mshych:
New reset_attr_ready file is created and set to 1 when all reset cause links
for this system are created. It can be used as a trigger in other services
that they can check reset reason.

Reviewed-by: Vadim Pasternak <vadimp@nvidia.com>
Signed-off-by: Michael Shych <mshych@nvidia.com>

### DIFF
--- a/usr/usr/bin/hw-management.sh
+++ b/usr/usr/bin/hw-management.sh
@@ -95,6 +95,7 @@ quantum2_pci_id=d2f2
 nv3_pci_id=1af1
 nv4_pci_id=22a3
 leakage_count=0
+reset_dflt_attr_num=18
 
 # Topology description and driver specification for ambient sensors and for
 # ASIC I2C driver per system class. Specific system class is obtained from DMI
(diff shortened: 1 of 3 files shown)


mshych:
RERUN

usr/usr/bin/hw-management.sh
```diff
@@ -95,6 +95,7 @@ quantum2_pci_id=d2f2
 nv3_pci_id=1af1
 nv4_pci_id=22a3
 leakage_count=0
+reset_dflt_attr_num=19
```
ciju-nvidia, line 98: How is the difference in number of reset attributes between 4.19 Vs 5.10 kernels are handled?

usr/usr/bin/hw-management.sh
```diff
@@ -878,6 +879,7 @@ msn274x_specific()
 	echo 2 > $config_path/cpld_num
 	echo 24c02 > $config_path/psu_eeprom_type
 	lm_sensors_config="$lm_sensors_configs_path/msn2740_sensors.conf"
+	echo 8 > $config_path/reset_attr_num
```
ciju-nvidia, line 882: This approach will not scale as and when we add/delete some reset causes. We will have to modify each and every *_specific() functions to accommodate the new/old one.

Wouldn't it be better if we have a table in place and in each *_specific() function just take the relevant entry?

Platform  |  4.19  |  5.10  |  6.1 |
---------------------------------
msn274x |    x     |   y      |   z   |  
--------------------------------
                 ......

PR #906, opened 2023-08-20
hw-mgmt: thermal: TC - disable autodetect for drivetemp sensor

sholeksandr:
TC - disable autodetect for drivetemp sensor


sholeksandr:
RERUN

yyehudai (approved):

PR #920, opened 2023-08-25
hw-mgmt: thermal: TC code improvement

sholeksandr:
1. Fix log sensor name for ambient sensor. Add failed file path instead
of just print sensor name

2. Fixed psu fan reading error.
If PSU not present no sense to add sensor_reading error for psu fan

3. Set pwm update worker immediate run only for hight PWM trend
It will make pwm_max_reduction timing more accurate in case several sensors
want to set pwm less pwm_max_reduction each but more than pwm_max_reduction in sum.


sholeksandr:
RERUN

vadimp-nvidia (commented):
Sasha, amend with s/hight/high
Set pwm update worker immediate run only for hight PWM trend

Add my tag: Reviewed-by: Vadim Pasternak <vadimp@nvidia.com>

sholeksandr:
RERUN

vadimp-nvidia (approved):

PR #921, opened 2023-08-27
hw-mgmt: system events: ASIC init complition done

vadimp-nvidia:
Add two new attributes:
/var/run/hw-management/config/asic_chipup_completed /var/run/hw-management/config/asics_init_done

'asic_chipup_completed' carries counter of completed ASIC devices probing.
'asics_init_done' is to be set to one, when 'asic_chipup_completed' attribute matches 'asic_num' attribute.

Motivation is to use 'asics_init_done' as criteria of system infrastructure readiness.

Reviwed-by: Felix Radensky <fradensky@nvidia.com>
Reviwed-by: Michael Shych <michaelsh@nvidia.com>

### DIFF
--- a/usr/usr/bin/hw-management-helpers.sh
+++ b/usr/usr/bin/hw-management-helpers.sh
@@ -242,6 +242,33 @@ change_file_counter()
 	echo $counter > $file_name
 }
 
+# Update counter, match attribute, unlock.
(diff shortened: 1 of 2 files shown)


vadimp-nvidia:
RERUN

felixradensky (approved):

PR #925, opened 2023-08-30
hw-mgmt: thermal: Fix PWM set 255 on ASIC disable

sholeksandr:
On systems with ASIC FAN control (SPC1), TC will not be able to change PWNM
after ASIC is disabled. To prevent system overheating, added set PWM
to maximum (255) when ASIC state changed to disabled.

Bug: #3586921

### DIFF
--- a/usr/usr/bin/hw-management.sh
+++ b/usr/usr/bin/hw-management.sh
@@ -2516,6 +2516,19 @@ do_chip_up_down()
 	case $action in
 	0)
 		lock_service_state_change
+
+		# If FAN PWM is controllded by ASIC - set it to 100%
+		pwm_link=$thermal_path/pwm1
+		if [ -L ${pwm_link} ] && [ -e ${pwm_link} ];
+		then
+			pwm_src=$(readlink -f ${pwm_link})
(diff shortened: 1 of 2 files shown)


sholeksandr:
RERUN

usr/usr/bin/hw-management.sh
```diff
@@ -2516,6 +2516,18 @@ do_chip_up_down()
...
 		lock_service_state_change
+
+		# If FAN PWM is controllded by ASIC - set it to 100%
+		pwm_link=$thermal_path/pwm1
+		if [ -L ${pwm_link} ] && [ -e ${pwm_link} ];
+		then
+			pwm_src=$(readlink -f ${pwm_link})
+			asic_i2c_add=/sys/devices/platform/mlxplat/i2c_mlxcpld.1/i2c-1/i2c-"$bus"/"$bus"-"$i2c_asic_addr_name"
+			if [[ "$pwm_src" == *"$asic_i2c_add"* ]]; then
+				echo  255 > $pwm_link
```
vadimp-nvidia, line 2527: Probably good to add some log, like:
log_info "Set PWM to maximum speed prior fan driver removing."
?

MellanoxBSP (approved):

PR #926, opened 2023-08-31
hw-mgmt: thermal: Fix PWM set 255 on ASIC disablei

sholeksandr:
On systems with ASIC FAN control (SPC1), TC will not be able to change PWNM
after ASIC is disabled. To prevent system overheating, added set PWM
to maximum (255) when ASIC state changed to disabled.

Bug: #3586921

Reviewed-by: Vadim Pasternak <vadimp@nvidia.com>


sholeksandr:
RERUN

sholeksandr:
RERUN

PR #927, opened 2023-08-31
hw-mgmt: thermal: Revert "Set pwm update worker immediate run only for high PWM trend"

sholeksandr:
1. Revert change ""Set pwm update worker immediate run only for high PWM trend"
2. Removed error reading of psu{X}_fan1_speed_get from error cases
TC not using psu1_fan1_speed_get value in PWM calculation. TC only shows
PSU FAN speed "just for information" in the log and should not be considered as thermal  error case.


sholeksandr:
RERUN

PR #936, opened 2023-09-11
hw-mgmt: scripts: Set pwm value to default on hw-management init

sholeksandr:
Set pwm value to default (20%) on hw-management init

### DIFF
--- a/usr/usr/bin/hw-management-thermal-events.sh
+++ b/usr/usr/bin/hw-management-thermal-events.sh
@@ -56,6 +56,7 @@ fan_drwr_num=0
 # 46 - F, 52 - R
 fan_direction_exhaust=46
 fan_direction_intake=52
+pwm_min_level=51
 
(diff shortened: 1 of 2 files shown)


sholeksandr:
RERUN

usr/usr/bin/hw-management-thermal-events.sh
```diff
@@ -323,6 +323,11 @@ if [ "$1" == "add" ]; then
 
 				if [ -f "$3""$4"/pwm1 ]; then
 					ln -sf  "$3""$4"/pwm1 "$tpath"/pwm1
+					pwm_level=$(< "$thermal_path/pwm1")
+					# If PWM level less then minimum (20%) set it to default value (20%)
+					if [ $pwm_level -lt 51 ]; then
```
ciju-nvidia, line 328: May be it's better to define it in a global variable and use the variable here. So, it will be easy to change it later if required.

ciju-nvidia (approved):

sholeksandr:
RERUN

PR #939, opened 2023-09-19
hw-mgmt: patches kernel 5.10: fix timing issue of add FAN EEPROMs on SN2201.

sholeksandr:
Define the exact i2c bus (adapter number) of fans on the SN2201 system.
This will cause fan's EEPROMs be connected already from nvsw-sn2201
platform driver and not from user space after receiving udev events.


sholeksandr:
RERUN

sholeksandr:
RERUN

sholeksandr:
RERUN

yyehudai (approved):

PR #945, opened 2023-09-21
Fix several issues with multi-asic systems

felixradensky:


felixradensky:
RERUN

usr/usr/bin/hw-management-thermal-events.sh
```diff
@@ -1088,9 +1086,6 @@ else
 			rm -f "$tpath/module*_temp_crit"
 			rm -f "$tpath/module*_temp_emergency"
 
-			echo 0 > $cpath/module_counter
```
vadimp-nvidia, line 1091: Why you remove setting to zero?
Do we remove $cpath" folder at all in this case?

usr/usr/bin/hw-management.sh
```diff
@@ -1536,7 +1536,7 @@ mqm9520_specific()
 		add_i2c_dynamic_bus_dev_connection_table "${mqm9520_dynamic_i2c_bus_connect_table[@]}"
 		add_cpu_board_to_connection_table $cpu_bus_offset
 	fi
-	i2c_asic2_bus_default=10
+	asic_i2c_buses=(2 10)
 	i2c_bus_def_off_eeprom_cpu=24
```
vadimp-nvidia, line 1540: Here it will be necessary to extend to vector of for for BM?

usr/usr/bin/hw-management-thermal-events.sh
felixradensky: Because on system with several ASICs when one ASIC is going down the module counter should not become 0

usr/usr/bin/hw-management.sh
felixradensky: Yes, I do that in the patch that adds XDR systems

vadimp-nvidia:
But you remove setting zero now for all systems?

PR #948, opened 2023-09-26
hw-mgmt: topology: Add helper function to detect ASIC hwmon existence

vadimp-nvidia:
Add helper function for ASIC hwmon detection.
On some systems, like SN2201 this process can take more time due to slower CPU.
So, make sure that hwmon obecect is exist in chipup flow.

### DIFF
--- a/usr/usr/bin/hw-management.sh
+++ b/usr/usr/bin/hw-management.sh
@@ -2606,6 +2606,15 @@ do_stop()
 	fi
 }
 
+function find_asic_hwmon_path()
+{
+	local path=$1
+	if [ ! -d "$path" ]; then
+		return 1
+	fi
+	return 0
+}
+
 do_chip_up_down()
 {
 	action=$1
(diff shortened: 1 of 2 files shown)


vadimp-nvidia:
RERUN

usr/usr/bin/hw-management.sh
```diff
@@ -2708,8 +2717,8 @@ do_chip_up_down()
 			set_i2c_bus_frequency_400KHz
 			echo mlxsw_minimal $i2c_asic_addr > /sys/bus/i2c/devices/i2c-"$bus"/new_device
 			restore_i2c_bus_frequency_default
-
-			if [ ! -d /sys/bus/i2c/devices/"$bus"-"$i2c_asic_addr_name"/hwmon ]; then
+			retry_helper find_asic_hwmon_path 0.2 3 "chip hwmon object" /sys/bus/i2c/devices/"$bus"-"$i2c_asic_addr_name"/hwmon
```
ciju-nvidia, line 2720: Is this 0.2 sec delay enough? Perhaps 1 sec, would be safe?

yyehudai (approved):

PR #954, opened 2023-10-01
hw-mgmt: thermal: remove depricated links coolingX_state

sholeksandr:
Remove depricated links:
coolingX_cur_state
coolingX_max_state

Since kernel thermal is disabled, we don't need these attributes and
those can be removed


sholeksandr:
RERUN

vadimp-nvidia (approved):

PR #972, opened 2023-10-16
hw-mgmt: thermal: Fix issue "RuntimeError: reentrant call inside"

sholeksandr:
Fix issue:  Logging Error in TC Log writing dmesg: RuntimeError: reentrant call inside

Bug: 3630148

### DIFF
--- a/usr/usr/bin/hw_management_thermal_control.py
+++ b/usr/usr/bin/hw_management_thermal_control.py
@@ -483,6 +483,7 @@ class Logger(object):
         self.logger.setLevel(logging.DEBUG)
         self.logger.propagate = False
         self.logger_fh = None
+        self.logger_emit = True
 
         self.set_param(use_syslog, log_file, verbosity)
 
@@ -536,6 +537,13 @@ class Logger(object):
         for handler in handler_list:
             handler.close()
             self.logger.removeHandler(handler)
+        self.logger_emit = False
+
+    def close_tc_log_handler(self):
+        if self.logger_fh:
+            self.logger_fh.flush()
+            self.logger_fh.close()
+            self.logger.removeHandler(self.logger_fh)
 
     def set_loglevel(self, verbosity):
         """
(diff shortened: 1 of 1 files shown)


sholeksandr:
RERUN

sholeksandr:
RERUN

usr/usr/bin/hw_management_thermal_control.py
```diff
@@ -2613,6 +2640,7 @@ def sig_handler(self, sig, *_):
                 self.stop(reason="SIG {}".format(sig))
 
             self.log.notice("Thermal control stopped", 1)
```
ciju-nvidia, line 2642: How does this resolve the problem when, TC was in side the logging function, and gets a signal? Signal handler will be calling self.log.notice and it would result in the reentrant RuntimeError.

sholeksandr: This will solve errer when we het logging error and then trying to log it from the logger:

    try:
        thermal_management = ThermalManagement(args, logger)
        thermal_management.init()
        thermal_management.start(reason="init")
        thermal_management.run()
    except BaseException as e:
        logger.info(traceback.format_exc())
        if thermal_management:
            thermal_management.stop(reason="crash ({})".format(str(e)))
            sys.exit(1)

sholeksandr: This will solve "reentrant" error wen we got logging error and trying log it form the main:
try:
    thermal_management = ThermalManagement(args, logger)
    thermal_management.init()
    thermal_management.start(reason="init")
    thermal_management.run()
except BaseException as e:
    logger.info(traceback.format_exc())
    if thermal_management:
        thermal_management.stop(reason="crash ({})".format(str(e)))
        sys.exit(1)

sholeksandr:
RERUN

ciju-nvidia (approved):

PR #976, opened 2023-10-17
Fix PSU attribute values for virtual platforms

ciju-nvidia:
While the platforms are running in virtual environment, there is some timing issue with loading the i2c traces for PSU and hence the initial reads of PSU attributes are wrong. This is a workaround to fix the issue by not changing the workflow for PSU initialization when creating PSU vpd.


ciju-nvidia:
RERUN

vadimp-nvidia (approved):

PR #986, opened 2023-10-23
hw-mgmt: thermal: TC add max threshold for PSU not present error

sholeksandr:
If psu_missing less or eq than psu_max_missing_err param - not classify it
as "psu_missing" error and any more PSU missing will be the 2nd error.


sholeksandr:
RERUN

vadimp-nvidia (approved):

PR #994, opened 2023-10-31
hw-mgmt: sensors: Fix PSU labels

felixradensky:
In multiple sensor.conf files PSU-1 and PSU-2 labels are swapped. PSU-1 label should match PSU at I2C address 0x59, while PSU-2 label should match PSU at I2C address 0x58. This patch corrects PSU labeling.


felixradensky:
RERUN

sholeksandr (approved):

PR #996, opened 2023-10-31
hw-mgmt: thermal: Fixed TC FAN DRWR number for MSN3700C system

sholeksandr:
Fixed FAN drwr count for msn3700C system. Was defined: 6, should be: 4

Bug: 3649678


sholeksandr:
RERUN

PR #1004, opened 2023-11-07
hw-mgmt: devtree: Add support for port board type

felixradensky:
1. Add new port board type
  2. Move port ambient sensor on XDR systems from main to port board


felixradensky:
RERUN

sholeksandr (approved):

PR #1007, opened 2023-11-08
hw-mgmt: scripts: fix fan direction setting on Panther-comex system.

mshych:
This is fix for bug# 3660884

Panther-comex is combined system, it has new CPLD fan control but fan direction
is still should be checked in old way from fan EEPROM.


mshych:
RERUN

vadimp-nvidia (approved):

mshych:
RERUN

PR #1015, opened 2023-11-16
hw-mgmt: udev events: Add extra parameter for ASIC core driver events

vadimp-nvidia:
Add environment parameter to udev event to distinct between ADD/DELETE events raised from 'sx_core' driver and same events raised because of PCI reset.
In the first case 'minimal' driver should be triggered, while in the second case not.

This additional parameter aims to filter out events raised by PCI reset.


vadimp-nvidia:
RERUN

ciju-nvidia (approved):

felixradensky (approved):

PR #1017, opened 2023-11-16
hw-mgmt: dump: Avoid file changed message

ciju-nvidia:
While running hw-management-generate-dump.sh, if the system is stressed and the file read operations are not completed within the given time interval, timeouts are defined to stop the ongoing read operations. This may generate a warning 'tar: ./hw-management_val: file changed as we read it' while the files are being archived by tar command. This patch increases the timeout intervals and adds a few second delay before starting the archieving operation.

Bug: 3450086


usr/usr/bin/hw-management-generate-dump.sh
```diff
@@ -120,6 +120,7 @@ dump_cmd "sensors" "sensors" "20"
 dump_cmd "iio_info" "iio_info" "5"
 dump_cmd "cat $REGMAP_FILE 2>/dev/null" "cpld_dump" "5"
 
-
+# Wait for a few seconds to complete all the timedout operations
+sleep 5
```
vadimp-nvidia, line 124: Ciju, why we need here 5 sec delay? Don't we have some criteria to check that two 'timeout's completed?

vadimp-nvidia (approved):

ciju-nvidia:
RERUN

PR #1023, opened 2023-11-19
hw-mgmt: patch table: Update patch table for v5.10

vadimp-nvidia:
Add to the list downstream patch #0098:
Use weak reverse dependencies for firmware flashing selection.


vadimp-nvidia:
RERUN

mshych (approved):

PR #1025, opened 2023-11-19
hw-mgmt: patches: 6.1: Add retry mechanism for failed transactions

vadimp-nvidia:
Sometimes I2C transactions could broken or non-completed becasue of some noise on I2C line or because ASIC resources is busy handling big amount of PCIe tarnsactions.

Add re-try mechanism for re-sending transaction which was not properly completed.
Retry up to three times and produce error log only in case the last try is not successful.


vadimp-nvidia:
RERUN

felixradensky (approved):

PR #1027, opened 2023-11-20
hw-mgmt: sensor: labels: New naming format

ciju-nvidia:
This patch set converts the sensor labels to new
naming format.


ciju-nvidia:
RERUN

vadimp-nvidia (approved):

PR #1037, opened 2023-11-28
hw-mgmt: scripts: Fix mlxsw_minimal driver init triggered by SDK udev…

felixradensky:
… event

When mlxsw_minimal driver initialization is triggered by Nvidia SDK udev event the value of ASIC index passed to hw-mgmt ASIC init code is 0. This special case was not handled correctly. Additionally, the case of legacy systems was not handled correctly by hw-mgmt routine mapping ASIC PCI bus ID to I2C bus ID.

Fixes: #3684822


felixradensky:
RERUN

vadimp-nvidia (approved):

PR #1041, opened 2023-11-30
hw-mgmt: scripts: Fix vpd parser to support DPU block

sholeksandr:
Fix vpd parser to support DPU block


sholeksandr:
RERUN

PR #1045, opened 2023-12-04
hw-mgmt: Add support for new system P4300

sholeksandr:
Add support for new system P4300


sholeksandr:
RERUN

usr/usr/bin/hw-management-global-wp.sh
```diff
@@ -187,7 +187,7 @@ release|restore)
 	echo "$global_wp_pid" > /var/run/hw-management-global-wp.pid
 	# systems without global WP but with ASIC WP
 	sku=$(< /sys/devices/virtual/dmi/id/product_sku)
-	if [ "$sku" == "HI142" ] || [ "$sku" == "HI152" ]; then
+	if [ "$sku" == "HI142" ] || [ "$sku" == "HI152" ] || [ "$sku" == "HI159" ]; then
```
vadimp-nvidia, line 190: I think we don't have Global WP support.
By the way it didn't work even on Vulcan.
Up to you - but you can drop Oberon from this case.

usr/usr/bin/hw-management-helpers.sh
```diff
@@ -157,13 +157,13 @@ find_i2c_bus()
...
-		default)
-			;;
-		esac
+                case $sku in
+                    HI151|HI156)
+                        i2c_bus_offset=$((i2c_bus_offset-1))
+                    ;;
+                        default)
+                    ;;
+                esac
```
vadimp-nvidia, line 166: What you changed here, tabs?
See that default) case is not aligned.

usr/usr/bin/hw-management.sh
```diff
@@ -516,13 +516,32 @@ qm3000_base_connect_table=( \
...
+	adt75 0x48 7 \
+	adt75 0x49 7 \
+	adt75 0x4a 7 \
+	adt75 0x4b 7 \
+	adt75 0x4c 7 \
+	adt75 0x4d 7 \
+	adt75 0x4e 7 \
+	adt75 0x4f 7 \
+	24c512 0x51 8 \
+	24c512 0x54 8 )
```
vadimp-nvidia, line 531: Do we need it having BOM SMBIOS?

PR #1047, opened 2023-12-04
hw-mgmt: Add support for new system P4300

sholeksandr:
Add support for new system P4300


sholeksandr:
RERUN

vadimp-nvidia (approved):

PR #1048, opened 2023-12-05
hw-mgmt: patches: Add temporary mux driver patch to WA HW problem

felixradensky:
Add delay to I2C mux driver channel selection routine to mitigate HW problem on XDR systems.


felixradensky:
RERUN

vadimp-nvidia (approved):

PR #1049, opened 2023-12-06
hw-mgmt: thermal: Fix error retcode in case of TC can't start

sholeksandr:
Fix error retcode in case of TC can't start


sholeksandr:
RERUN

vadimp-nvidia (approved):

PR #1055, opened 2023-12-10
hw-mgmt: dump: Move output of sensors to the dump start

vadimp-nvidia:
Perform dump of sensors prior to journal dump.


vadimp-nvidia:
RERUN

mshych (approved):

PR #1058, opened 2023-12-11
hw-mgmt: scripts: Add system_flow_capability attribute for XDR systems

felixradensky:
XDR systems support only reverse fan direction, system_flow_capability attribute is required by TC on systems with single flow direction.


felixradensky:
RERUN

sholeksandr (approved):

yyehudai (approved):

PR #1061, opened 2023-12-12
hw-mgmt: scripts: Update SimX supported platforms

ciju-nvidia:
This patch set adds more platforms to SimX support. Also it re-organizes the check_simx() function to take the list of platforms from a single variable. It also enables the TC explicitly for SimX platforms.

Bug #3696439

### DIFF
--- a/usr/usr/bin/hw-management-ready.sh
+++ b/usr/usr/bin/hw-management-ready.sh
@@ -1,6 +1,6 @@
-#!/bin/sh
+#!/bin/bash
 ##################################################################################
-# Copyright (c) 2020 - 2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright (c) 2020 - 2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions are met:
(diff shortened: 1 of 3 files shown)


ciju-nvidia:
RERUN

ciju-nvidia:
RERUN

usr/usr/bin/hw-management-ready.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 ##################################################################################
 # Copyright (c) 2020 - 2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
```
vadimp-nvidia, line 4: Need to update license date. I see we didn't change it since 2021.

usr/usr/bin/hw-management-ready.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
vadimp-nvidia, line 1: What we lack in sh shell?

usr/usr/bin/hw-management-ready.sh
ciju-nvidia: Updated as 2020 - 2023

usr/usr/bin/hw-management-ready.sh
ciju-nvidia: I wanted to 'source hw-management-helpers.sh'. So changed to 'bash'
Do you want  me to do '. hw-management-helpers.sh' instead and retain '/bin/sh'?

ciju-nvidia:
RERUN

sholeksandr (approved):

PR #1067, opened 2023-12-13
hw-mgmt: thermal: TC fix crash on FAN disconnect.

sholeksandr:
TC fix crash on FAN disconnect. This fix for code which
was added in previous commit


felixradensky (approved):

sholeksandr:
RERUN

PR #1070, opened 2023-12-13
hw-mgmt: scripts: Fix sku variable for SimX check

ciju-nvidia:
This patch fixes the wrong variable which was used for checking the platform sku when running in the SimX environment.


ciju-nvidia:
RERUN

vadimp-nvidia (approved):

felixradensky (approved):

PR #1073, opened 2023-12-18
hw-mgmt: thermal: TC use tc_config copy instred of soft link

sholeksandr:
TC use tc_config copy instred of soft link


sholeksandr:
RERUN

PR #294, opened 2021-12-17
hw-mgmt: scripts cleanup optimization

sholeksandr:
scripts cleanup and optimization

Signed-off-by: Oleksandr Shamray <oleksandrs@nvidia.com>

PR #1077, opened 2023-12-19
hw-mgmt: pathes (sonic): v5.10: Modify patch for mp2891 device

felixradensky:
Update patch #0294 (opt and sonic).


felixradensky:
RERUN

vadimp-nvidia (approved):

PR #1078, opened 2023-12-19
hw-mgmt: scripts: Fix asic health attribute

ciju-nvidia:
In SimX environment asic_health is 3. This will cause the minimal driver not to load. This patch will probe minimal always and run chipup only for those with good health.


ciju-nvidia:
RERUN

felixradensky (approved):

vadimp-nvidia (approved):

PR #1079, opened 2023-12-20
hw-mgmt: scripts: Increase the max number of PSUs and tachos

felixradensky:
On XDR systems the max number of PSUs is 8, max number of tachos is 20.


mshych (approved):

felixradensky:
RERUN

PR #1085, opened 2023-12-31
hw-mgmt: patches: 5.10: Add temporary mlxsw_minimal driver patch

felixradensky:
Add temporary mlxsw_minimal driver patch to disable ethtool interface


felixradensky:
RERUN

PR #1100, opened 2024-01-08
hw-mgmt: kernel: kconfig: Set CONFIG_IGB=y and CONFIG_E1000E=y

felixradensky:
This sets the correct MGMT interface initialization order on systems with 2 MGMT interfaces.


felixradensky:
RERUN

recipes-kernel/linux/kconfig.txt
```diff
@@ -62,8 +62,9 @@ CONFIG_CPU_THERMAL=y
 CONFIG_X86_PKG_TEMP_THERMAL=m
 CONFIG_SENSORS_CORETEMP=m
 CONFIG_INTEL_PCH_THERMAL=m
-CONFIG_IGB=m
+CONFIG_IGB=y
```
ciju-nvidia, line 65: We had made 'CONFIG_IGB=m' based on the feedback from SONiC team as it was causing the build failures in SONiC. Can't we retain it as a module? Any specific reason to make it part of the kernel?

PR #1102, opened 2024-01-08
hw-mgmt: scripts: Prevent udev handling before hw-mgmt basic init is …

felixradensky:
…done

Make sure to always exit udev handling scripts before hw-management is ready to handle udev events.


felixradensky:
RERUN

ciju-nvidia (approved):

PR #1103, opened 2024-01-08
hw-mgmt: scripts: ui: Fix naming for sensors

ciju-nvidia:
Correct the names for temperature sensors in the json file to match with the new naming format.


ciju-nvidia:
RERUN

PR #1107, opened 2024-01-10
hw-mgmt: ui: Fix issue with fan ambient sensor

ciju-nvidia:
Update the HI157 ui archive to fix the fan ambient sensor name.


ciju-nvidia:
RERUN

felixradensky (approved):

PR #1110, opened 2024-01-10
hw-mgmt: sensors: Update sensor configs for XDR systems

felixradensky:
Ignore temp2 sensors on MP2891 VRs where these sensors are not connected.


felixradensky:
RERUN

mshych (approved):

PR #1112, opened 2024-01-10
hw-mgmt: scripts: Add generation of asic_bus attribute on multi-ASIC …

felixradensky:
…systems

This leagacy attribute is required for backward compatibility.


felixradensky:
RERUN

yyehudai (approved):

PR #1119, opened 2024-01-15
hw-mgmt: scripts: add DPU and Comex AMD to SMBIOS BOM plus improvements.

mshych:
Add BF3 DPU to SMBIOS BOM.
Add AMD Comex to SMBIOS BOM.
Generalization of multiple boards processing.
Code clean.

### DIFF
--- a/usr/usr/bin/hw-management-devtree.sh
+++ b/usr/usr/bin/hw-management-devtree.sh
@@ -35,13 +35,13 @@ devtr_verb_display=0
 devtree_codes_file=
 
 # Declare common associative arrays for SMBIOS System Version parsing.
-declare -A board_arr=(["C"]="cpu_board" ["S"]="switch_board" ["F"]="fan_board" ["P"]="power_board" ["L"]="platform_board" ["K"]="clock_board" ["O"]="port_board")
+declare -A board_arr=(["C"]="cpu_board" ["S"]="switch_board" ["F"]="fan_board" ["P"]="power_board" ["L"]="platform_board" ["K"]="clock_board" ["O"]="port_board" ["D"]="dpu_board")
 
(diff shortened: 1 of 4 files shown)


usr/usr/bin/hw-management-devtree.sh
```diff
@@ -542,8 +583,28 @@ devtr_check_supported_system_init_alternatives()
 			done
 			return 0
 			;;
+		VMOD0019)
+			case $sku in
+				HI160)
```
vadimp-nvidia, line 588: You have redundant tabs here.

usr/usr/bin/hw-management-devtree.sh
```diff
@@ -552,11 +613,10 @@ devtr_check_supported_system_init_alternatives()
...
 	local board_str=$1
-	local clk_board_num=1
-	local clk_bus_offset=0
-	local clk_addr_offset=0
-	local swb_board_num=1
-	local swb_bus_offset=0
+	local board_num=1
+	local board_bus_offset=0
+	local board_name_pfx=
+#	local board_addr_offset=0
```
vadimp-nvidia, line 619: It seems leftover comment for board_addr_offset

mshych: No, It's not leftover but I specially left it commented as an example.
There is also a comment:
# There is no currently address offset. Leave commented just for possible future use

usr/usr/bin/hw-management-devtree.sh
mshych: Strange, I checked and see it correct both in file itself and in github diff
diff

mshych: It's the same as in all function, other cases:
diff

vadimp-nvidia (approved):

PR #1121, opened 2024-01-16
hw-mgmt: thermal: Add list of supported PSU PN to extract fan direction.

sholeksandr:
Add list of supported PSU PN to extract fan direction.


sholeksandr:
RERUN

felixradensky (approved):

PR #1130, opened 2024-01-17
hw-mgmt: scripts; Add device probe retry support

felixradensky:
Make several attempts to probe for I2C devices if
initial probe fails.

### DIFF
--- a/usr/usr/bin/hw-management.sh
+++ b/usr/usr/bin/hw-management.sh
@@ -108,6 +108,7 @@ dpu_bf3_pci_id=c2d5
 leakage_count=0
 leakage_rope_count=0
 asic_chipup_retry=2
+device_connect_retry=2
 chipup_log_size=4096
 reset_dflt_attr_num=18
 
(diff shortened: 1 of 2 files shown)


felixradensky:
RERUN

vadimp-nvidia (approved):

usr/usr/bin/hw-management.sh
```diff
@@ -2428,8 +2429,14 @@ connect_platform()
 	fi
 
 	for ((i=0; i<${#connect_table[@]}; i+=$dev_step)); do
-		connect_device "${connect_table[i]}" "${connect_table[i+1]}" \
-				"${connect_table[i+2]}"
+		for ((j=0; j<${device_connect_retry}; j++)); do
```
ciju-nvidia, line 2432: We already have a retry_helper() routine in 'hw-management-helpers.sh'. Could we make use of this instead of having custom loops?

felixradensky: I don't think that using retry_helper() is good in this case. It will not improve readability or make the code smaller.

PR #1139, opened 2024-01-25
Revert "hw-mgmt: scripts; Add device probe retry support"

felixradensky:
This reverts commit ab9f661b97f40bacebe0db1edbcc43c1b3553281.

### DIFF
--- a/usr/usr/bin/hw-management-helpers.sh
+++ b/usr/usr/bin/hw-management-helpers.sh
@@ -118,7 +118,6 @@ lc_i2c_bus_min=34
 lc_i2c_bus_max=43
 i2c_bus_offset=0
 cpu_type=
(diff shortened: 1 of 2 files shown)


felixradensky:
RERUN

mshych:
Why do you revert this retry? What's wrong?
I just use a similar mechanism for dynamic boards/devices.
Should I remove retries also there?

felixradensky:
This mechanism was added to fix the probing failure of fan ambient sensor on BF3 Leopard. It broke MP2891 probing on BM Surrogate, the 0.2 sec delay between probe and test is too small, I've tried 0.5sec and it also didn't help. Increasing the delay even more will negatively affect boot time, as it's done for each device.

PR #1157, opened 2024-02-15
hw-mgmt: scripts: Fix hotplug events for dpu

ciju-nvidia:
DPU power/health events were not updated in the
/var/run/hw-management/dpu*/events folder. It was due to the fact that attributes did not have the dpu slot number in the attribute name. This patch introduces a new tag 'hotplug-dpu-event' and an action function to find out the dpu slot number, which is used to derive the event attribute path.


vadimp-nvidia (approved):

ciju-nvidia:
RERUN

PR #1159, opened 2024-02-16
hw-mgmt: scripts: Fix dpu_shtdn action script

ciju-nvidia:
The pattern dpu*_ready in hotplug action script was causing both dpu[1-8]_ready and dpu[1-8]_shtdn_ready events to execute the dynamic sensor loading function. The problem was that the wild card character '*' was redirecting dpu*_shtdn_ready events to dpu*_ready case, because it ends with '_ready'. This patch fixes this problem by having the dpu slot numbers in the pattern checking, there by avoiding the '*' wild card character.


ciju-nvidia:
RERUN

sholeksandr (approved):

PR #1161, opened 2024-02-19
hw-mgmt: scripts: Check for dpu events

ciju-nvidia:
Introducing a check for dpu_ready and dpu_shtdn_ready events to see if the udev events are valid and then only
connect / disconnect dynamic sensors.


ciju-nvidia:
RERUN

vadimp-nvidia (approved):

PR #1165, opened 2024-02-22
hw-mgmt: scripts: Fix udev event path for sodimm

ciju-nvidia:
In virtual machines based on qemu, the jc42 driver attributes are located in a different path. It is not in the path /sys/devices/pci0000:00. So the sodimm sensor attributes were not created in /var/run/hw-management/thermal directory. This patch fixes this issue by adding a udev rule with the correct path for jc42 attributes on virtual platforms.


ciju-nvidia:
RERUN

ciju-nvidia:
RERUN

vadimp-nvidia (approved):

PR #1170, opened 2024-02-26
hw-mgmt: scripts: Mapping pci bus number to dpu slot

ciju-nvidia:
This patch introduces the pci bus to dpu slot number mapping. PCI buses can be different for CFL and AMD cpu comex. This also introduces a new config variable 'dpu_detected_num' which will tell the number of DPUs detected by lspci command.


vadimp-nvidia (approved):

ciju-nvidia:
RERUN

PR #1171, opened 2024-02-28
V.7.0030.4000 BR

mshych:
- hw-mgmt: udev: Change trigger rules for SODIMMs temp sensors.
- hw-mgmt: sensors: Add sensorss config rule for 2nd PSU in sn5600/sn5400 systems.


mshych:
RERUN

PR #1180, opened 2024-03-05
hw-mgmt: scripts Fix FAN dir attribute for PN in none-MLNX format

sholeksandr:
For the systems, with FANs equiped with FRU EEPRM fan_direction attr
value can be extracted from FRU PN field. If PN is not in MLNX
format (MTEF-FANR-A) or broken - not possible to get fan_direction
and fanX_dir attribute not crating.

With this fix added fan direction value == 2, which means "FAN
direction can't be extracted"


sholeksandr:
RERUN

sholeksandr:
RERUN

vadimp-nvidia (approved):

sholeksandr:
RERUN

sholeksandr:
RERUN

PR #1184, opened 2024-03-07
Disable 'ethtool' interface,

yyehudai:
Disable 'ethtool' interface, since it is un-used and just created un-necessary 'netdevice' interfaces.


yyehudai:
RERUN

vadimp-nvidia (approved):

PR #1196, opened 2024-03-26
hw-mgmt: infra: Increase time for validation chip hwmon object creation

vadimp-nvidia:
Increase for SP1/SPC2 systems only.

Fix: #3748535.


vadimp-nvidia:
RERUN

ciju-nvidia (approved):

PR #1216, opened 2024-04-17
hw-mgmt: scripts: VPD parser add support for IPMI fru

sholeksandr:
Add FRU decode for IPMI format. It requires ipmi-fru package to
be installed.


sholeksandr:
RERUN

PR #1218, opened 2024-04-18
hw-mgmt: patches: Add downstream mlxsw minimal driver patch

felixradensky:
Add downsteam mlxsw minimal driver patch to hardcode the number of modules per ASICs on QM3200 systems instead of reading the actual number of modules from the MGPIR register. This is a WA for recent FW versions that incorrectly report the number of modules per ASIC.


felixradensky:
RERUN

ciju-nvidia (approved):

PR #1226, opened 2024-05-02
V.7.0030.4000 BR

sholeksandr:
- hw-mgmt: patches: Add nvos type for --os argument patch_deploy_tool
- hw-mgmt: patches: Add dvs type for --os_type argument patch_deploy_tool


sholeksandr:
RERUN

PR #1237, opened 2024-05-17
hw-mgmt: thermal: TC fixed amb sensor handling for vaious sensor count

sholeksandr:
Most systems have 2 amb thermal sensors. But some can have more or less.
In this case TC amb sensor should handle only existing (predefined in TC config) phisical
amb_{X} sensors.

This commit add fix for TC abm thermal sensor PWM calculation for systems with amb_{X} count  != 2


sholeksandr:
RERUN

PR #1240, opened 2024-05-20
hw-mgmt: scripts: Rename qm3400 system to qm3200

sholeksandr:
Rename qm3400 system to qm3200

Bug: 3882472


felixradensky (commented):
The correct name is Q3200

PR #1258, opened 2024-06-07
hw-mgmt: thermal: Correcting the dpu voltmon attribute name

ciju-nvidia:
This patch changes the naming convention by removing the 'dpu' string from voltmon temperature attribues.


ciju-nvidia:
RERUN

sholeksandr (approved):

PR #1268, opened 2024-06-16
hw-mgmt: topology: n5110ld: Fix hotplug_fans

sholeksandr:
In N5110-LD statem, FANs are not hot plug. Value in otplug_fans attribute
should be 0 (zero).


sholeksandr:
RERUN

vadimp-nvidia (approved):

PR #1275, opened 2024-06-24
hw-mgmt: scripts: Add support for PSID block to vpd parser

sholeksandr:
Add to vpd parser support for up PSIID block.
This block contain string record up to 32 bytes length


sholeksandr:
RERUN

mshych (approved):

PR #1282, opened 2024-06-27
hw-mgmt: thermal: TC: disable auto add asic sensors to monitoring

sholeksandr:
Disable auto add asic sensors to monitoring. Add ASIC thermal sensors only
if it defined in TC thermal data .json


sholeksandr:
RERUN

PR #1290, opened 2024-07-01
hw-mgmt: topology: Add n5100 support

sholeksandr:
Add n5100 support HW_ID:167, SKU:HI167


sholeksandr:
RERUN

sholeksandr:
RERUN

sholeksandr:
RERUN

sholeksandr:
RERUN

usr/usr/bin/hw-management.sh
```diff
@@ -2246,62 +2246,33 @@ n5110ld_specific()
...
+			max_tachos=8
+			echo 4 > $config_path/fan_drwr_num
+			thermal_control_config="$thermal_control_configs_path/tc_config_n5110ld_ttm.json"
+		;;
+		HI167)	# NSO
+			max_tachos=8
+			echo 4 > $config_path/fan_drwr_num
+			thermal_control_config="$thermal_control_configs_path/tc_config_n5100ld.json"
+		;;
+		*)
```
mshych, line 2266: Would we like to have default Power-On system and not TTM?

sholeksandr: It was the first version. Usually we set default first version of the system. In any case it's case if SKU is not programmed. It shouldn't happen in real system.

felixradensky (approved):

PR #1293, opened 2024-07-09
hw-mgmt: patches: Update SN4280 platform driver

ciju-nvidia:
This patch removes the reset_from_carrier reset reason as it is assserted for every power cycle and watchdog reboots.

Bug #3981606


ciju-nvidia:
RERUN

vadimp-nvidia (approved):

PR #1294, opened 2024-07-09
hw-mgmt: thermal: Correct DPU voltmon sensor attribute

ciju-nvidia:
TC was expecting the voltmon temperature sensor attribute from dpu as dpu_voltmon[1-2]_temp1_input. This commit removes the 'dpu' prefix.


ciju-nvidia:
RERUN

felixradensky (approved):

PR #1299, opened 2024-07-15
Update SN4280 platform driver

ciju-nvidia:
1) This patch removes the reset_from_carrier reset reason as it is asserted for every power cycle and watchdog reboots.
    Bug #3981606

2) Correct DPU voltmon sensor attribute

3) Update the reset attribute count for SN4280


ciju-nvidia:
RERUN

felixradensky (approved):

PR #1300, opened 2024-07-16
hw-mgmt: thermal: Set PSU fan dir to "Unknown" if PSU FRU is broken

sholeksandr:
Set PSU fan dir value to "Unknown" (value 2) if PSU FRU is broken


sholeksandr:
RERUN

ciju-nvidia (approved):

PR #1310, opened 2024-07-23
hw-mgmt: patches: kernel 5.10/6.1: Fix fan count for N51XX platform

sholeksandr:
Fix fan count for N51XX platform


sholeksandr:
RERUN

vadimp-nvidia (approved):

PR #1312, opened 2024-07-25
hw-mgmt: scripts: Update attribute sync service

sholeksandr:
Revert mistakenly deleted asic1 thermal attribute


sholeksandr:
RERUN

PR #1315, opened 2024-07-26
hw-mgmt: scripts: Fix invalid dpu entry

ciju-nvidia:
Due to the usage of wildcards in udev rules, the dpu directories were getting created on most of the platforms where DPUs are not present. This patch fixes it by having a check for hw sku, before creating the dpu directories. There by limiting it to only DPU platforms.

The problematic udev rule is with DPU topology creation DEVPATH=="/devices/platform/mlxplat/i2c_mlxcpld.*/i2c-*/i2c-*/*-0068"

This one gets triggered for voltage monitors with address 0x68. Having a sku check will prevent it.

Bugs #4001444


ciju-nvidia:
RERUN

vadimp-nvidia (approved):

PR #1317, opened 2024-07-29
hw-mgmt: patches: Correct mp2891 voltage scaling

ciju-nvidia:
This patch fixes the issue while shifting the bits in pmbus register 0xBD. A right shift of 14 bits
would make the value as 0 and the check for BIT(13) will always fail. This was causing the wrong scale value of 5mv to be used (instead of 2.5mv) for vou2.

Bug #3962952


ciju-nvidia:
RERUN

vadimp-nvidia (approved):

ciju-nvidia:
RERUN

PR #1321, opened 2024-08-01
w-mgmt: scripts: Add fanX_speed_tolerance attribute

sholeksandr:
The fanX_speed_tolerance attribute is defining how mutch FAN RPM can be different
from expected max/min. Value defined in percent.
Example:

fan1_max = 10000
fan1_speed_tolerance = 15

In this case, range 8500..12000 is considered as normal.


sholeksandr:
RERUN

ciju-nvidia (approved):

PR #1344, opened 2024-08-14
hw-mgmt: thermal: TC fix crash for unsuppotred platform

sholeksandr:
In case hw-mgmgt running on unknown platform - TC should
treat it as "Unsupported platform".

Bug 4032775


sholeksandr:
RERUN

PR #1345, opened 2024-08-15
hw-mgmt: scripts: Fix init error fox SimX paltform

sholeksandr:
Fix SKU/platform reading if it's not exists. This can be when running on
SimX platform.

log messages example before fix:

/usr/bin/hw-management.sh: line 61: /sys/devices/virtual/dmi/id/board_name: No such file or directory
/usr/bin/hw-management.sh: line 62: /sys/devices/virtual/dmi/id/product_sku: No such file or directory
/usr/bin/hw-management-helpers.sh: line 325: /var/run/hw-management/events/: Is a directory

### DIFF
--- a/usr/usr/bin/hw-management-ready.sh
+++ b/usr/usr/bin/hw-management-ready.sh
@@ -40,8 +40,8 @@
 #              Report start of hw-management service to console and logger.
 
 source hw-management-helpers.sh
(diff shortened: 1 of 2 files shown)


sholeksandr:
RERUN

felixradensky (approved):

PR #1349, opened 2024-08-21
hw-mgmt: thermal: TC fix timer for while loops based on current system time

sholeksandr:
In while loops like:

  end = current_milli_time() + delay
  while end > current_milli_time():
    do_somthing()

We potentially can get incorrect delay in case if during loop system
time will be changed. It happens because current_milli_time() based on system time
and this time can be forwarded or rewinded.  To fix this issue - it needs to
change current_milli_time to time which can't be changed, like: system uptime,
application uptime etc.

This fix changing current_milli_time() to read CLOCK_PROCESS_CPUTIME_ID
(High-resolution per-process timer from the CPU) timer as reference.

Bug #4042294

### DIFF
--- a/usr/usr/bin/hw_management_thermal_control.py
+++ b/usr/usr/bin/hw_management_thermal_control.py
@@ -393,7 +393,7 @@ def current_milli_time():
         get current time in milliseconds
     @return: int value time in milliseconds
     """
-    return round(time.time() * 1000)
+    return round(time.clock_gettime(1) * 1000)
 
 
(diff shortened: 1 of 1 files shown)


sholeksandr:
RERUN

ciju-nvidia (approved):

ciju-nvidia:
RERUN

sholeksandr:
RERUN

sholeksandr:
RERUN

PR #1351, opened 2024-08-23
hw-mgmt: scripts: Fix coretemp module insert  on AMD systems

sholeksandr:
Fix issue "Failed to insert module 'coretemp': No such device".
Module 'coretemp' doesn't support AND CPU.

Bug: 4042338


sholeksandr:
RERUN

ciju-nvidia (approved):

PR #1375, opened 2024-09-06
V.7.0040.2000 BR

sholeksandr:
- hw-mgmt: thermal: Fix min/max range comparison for amb thermals sensor
- Add kernel 5.10.218 support + rebase BF3 patches


sholeksandr:
RERUN

felixradensky (commented):
In patch status table I suggest marking BF3 patches in the same manner for kernel 5.10 and 6.1 and use the string "BF3" in last column. Also, for kernel 6.1 BF3 patches are skipped for sonic and cumulus, I suggest doing the same for kernel 5.10

PR #1388, opened 2024-09-25
V.7.0040.2000 BR

sholeksandr:
- hw-mgmt: scripts: Fix hw-management-sync crash
- hw-mgmt: scripts: Fix power events init


sholeksandr:
RERUN

PR #1391, opened 2024-09-26
hw-mgmt: patches: Fixing mp2891 driver loading

ciju-nvidia:
In 5.10 kernel, mp2891 driver was not initialized. It was due to a corrupted patch. Fixing the patch.

Bugs #4048801 #4082431


ciju-nvidia:
RERUN

vadimp-nvidia (approved):

PR #1393, opened 2024-09-30
hw-mgmt: scripts: Fix power coretamp module load on SimX

sholeksandr:
Since SimX doesn't support CPU core temp emulation - disables 'coretemp'
module load in case of running on SimX

Bug: 4093394


sholeksandr:
RERUN

vadimp-nvidia (approved):

PR #1397, opened 2024-10-10
hw-mgmt: kernel: 6.1: Fix for PSU attributes

ciju-nvidia:
Starting from kernel 5.12.rc-1, pmbus reads are always forced down to chip and get the values directly. Driver doesn't cache the values anymore.

This behviour change has caused problems in Delta-1.1K PSUs, on which pmbus registers were not written.

This patch introduces the driver caching.

Bugs# 3874682


ciju-nvidia:
RERUN

vadimp-nvidia (approved):

PR #1400, opened 2024-10-15
hw-mgmt: patches: kernel 5.10/6.1: N51xx fix for power button reset cause

sholeksandr:
Fix for power button reset cause – reset cause should be
“reset_pwr_button_or_leak_con”.

Required fix in CPU - renaming:
“reset_leak_con” -> “reset_pwr_button_or_leak_con”.

Until we don’t have this fix, reset cause will be exposed as “reset_leak_con”.


sholeksandr:
RERUN

sholeksandr:
RERUN

PR #1402, opened 2024-10-24
V.7.0040.2000 BR

sholeksandr:
- hw-mgmt: patches: fix patch deploy tool
- hw-mgmt: patches: add kernel 5.14 support


sholeksandr:
RERUN

PR #1408, opened 2024-10-28
hw-mgmt: patches: 6.1: Add support for SN5640 sku HI172

sholeksandr:
1. Add kernel support for SN5640 sku:HI172
2. Fix ASIC count for SN5640


sholeksandr:
RERUN

PR #1413, opened 2024-11-04
hw-mgmt: sensors: Fix temperature sensor descriptions for SN5600

felixradensky:
SN5600 uses reverse (C2P) fans, so fan side temperature sensor is at air exhaust side and port side sensor is at air intake side.


felixradensky:
RERUN

acoifmannvidia (approved):

PR #1429, opened 2024-11-18
hw-mgmt: patches: Update mp2891 threshold limits

sholeksandr:
Fixing the wrong calculation in mp 2891 for VOUT thresholds. Remove
thresholds scaling by 1.05/0.95.


sholeksandr:
RERUN

PR #1431, opened 2024-11-18
hw-mgmt: patches 6.1: Add patch to fix uninitialized variable

vadimp-nvidia:
mlxsw: core: Downstream: Fix uninitialized variable

Set variable "status" to zero in function mlxsw_cmd_exec() to prevent garbage value to be passed from underling functions.

In case of transaction failure due to some hardware problem, uninitialized 'status' may be returned with some garbage value. This garbage can be interpreted as some special status, like MLXSW_CMD_STATUS_FW_ISSU (0x27) or MLXSW_CMD_STATUS_RUNNING_RESET (0x26).
Such codes are considered by the driver as a special flow, which is ended up with driver removing.
When such code is received - all the next transactions to ASIC are blocked.

In case this code is received unintentionally, driver will stop control PWM and stop monitor ASIC and receivers’ sensors and system may end up with thermal shutdown.


sholeksandr (approved):

ciju-nvidia (approved):

vadimp-nvidia:
RERUN

PR #1437, opened 2024-11-20
hw-mgmt: scripts: fix init errors of GPIOs for JTAG on AMD based systems.

mshych:
CPLD update on AMD-based systems is done through the LPC CPLD interface.
So, GPIOs for JTAG are not relevant and shouldn't be initialised.

Bug: 4116810


felixradensky (approved):

ciju-nvidia (approved):

ciju-nvidia:
RERUN

PR #1440, opened 2024-11-20
hw-mgmt: devtree: Extend support for Based Management Borad device ta…

vadimp-nvidia:
…ble parsing

Add device tree records explicitly for BMC CPU type, since BMC does not have DMI PN value, used for selection of device tree records on the system supporting DMI.

Replace shell command fold with shell implementation since it does not work in busybox.


vadimp-nvidia:
RERUN

mshych (approved):

PR #1446, opened 2024-11-21
hw-mgmt: scripts: Update dpu thermal script

ciju-nvidia:
This commit does the following:
 - Correct the dpu_num path
 - Copy the dpu thermal script to /var/run/hw-management/bin/

Bugs #417374 #4170646


ciju-nvidia:
RERUN

vadimp-nvidia (approved):

PR #1461, opened 2024-12-02
hw-mgmt: scripts: Fix ASIC temp populate in hw-mgmt-sync script

sholeksandr:
Fix hw-management sync crash in case ASIC value is not exist

Bug: 4184342


sholeksandr:
RERUN

vadimp-nvidia (approved):

sholeksandr:
RERUN

PR #1469, opened 2024-12-05
hw-mgmt: attributes: Fix max and crit values for ASIC temperatures

sholeksandr:
Fix in hw-mgmt sunc script max and crit values for ASIC temperatures

Bug: 4173495


sholeksandr:
RERUN

sholeksandr:
RERUN

felixradensky (approved):

PR #1472, opened 2024-12-05
hw-mgmt: config: Add Thermal config for SN5610

sholeksandr:
Add Thermal config for SN5610

Bug: 4200580


sholeksandr:
RERUN

PR #1477, opened 2024-12-06
hw-mgmt: attributes: Fix PSU volt out attributebute

sholeksandr:
On some PSUs volt out value exposed in out2 pmbus attribute (instead of
vout) With this fix - added correct vout attribute linkage on PSUs which
have both vout/vout2 attributes

Bug: 4143199


sholeksandr:
RERUN

vadimp-nvidia (approved):

PR #1481, opened 2024-12-09
hw-mgmt: scripts: Fix SODIMM naming on SN5610 and SN5640

felixradensky:
There is only one SODIMM module on SN5610 and SN5640 systems at I2C address 0x1b. Name it SODIMM1 and adjust all relevant files.


felixradensky:
RERUN

usr/usr/bin/hw-management-thermal-events.sh
mshych, line 1: I think you can revert 0x1b and 0x1a in all cases

felixradensky: This will not work for Juliet, which expects SODIMM1 at 0x1a

mshych (approved):

PR #1484, opened 2024-12-10
hw-mgmt: scripts: Fix module reading in sync script

sholeksandr:
Update the hw_management sync script to fix module temperature reading
when in SW mode. Key changes:

1. Add a check for control mode (SW or FW) before reading. Skip reading
in SW mode.
2. Add validation for module presence.
3. Run scan and update for all modules in a single loop instead of
updating each module separately.

Bug: 4201043


sholeksandr:
RERUN

felixradensky (approved):

PR #1487, opened 2024-12-11
hw-mgmt: scripts: Fix asic_chipup_completed for systems without minimal driver

sholeksandr:
Fix asic_chipup_completed for systems without minimal driver. Added
feature to sync script which will handle asic state and update
attributes/counters:

- asics_init_done
- asic_chipup_completed

Bug: 34304249


sholeksandr:
RERUN

PR #1488, opened 2024-12-11
hw-mgmt: rules: Add rules for network interface

felixradensky:
Add rules for plugin/plugout USB net interface connecting CPU host and BMC.

Bug: 4146744


felixradensky:
RERUN

ciju-nvidia (approved):

PR #1495, opened 2024-12-15
hw-mgmt: patches: Disable reset COME causes to avoid duplication

vadimp-nvidia:
Disable the below COME reset cuases, since they are coming as an additional 2-nd cause: "reset_sw_reset", "reset_from_carrier", "reset_aux_pwr_or_reload", "reset_platform", "reset_pwr".

Thus, two causes will be reported, while expectation is to provide only one.

Disable reset COME causes to avoid duplication.


vadimp-nvidia:
RERUN

sholeksandr (approved):

PR #1507, opened 2024-12-23
hw-mgmt: scripts: Add ASIC state check on temperature reading

sholeksandr:
When the ASIC SX driver is not available due to the ISSU state,
temperature reading is not available. This commit adds a check for the
asic{}_ready flag to determine if temperature reading is available:

/var/run/hw-management/config/asic{}_ready

1 - ASIC sx driver ready to temperature read
0 - ASIC driver not loaded of temperature is not avalible


sholeksandr:
RERUN

felixradensky (approved):

PR #1510, opened 2024-12-23
hw-mgmt: config: Fix config files creation order on hw-mgmgmt start

sholeksandr:
Fix config file creation order on hw-mgmt start

Bug: 4209209


felixradensky (approved):

sholeksandr:
RERUN

PR #1513, opened 2024-12-23
hw-mgmt: thermal: TC add check for abnmormal values on termal sensors

sholeksandr:
TC add check for abnmormal values on termal sensors.

For example:

ASIC temperature can't be 0 (Possible switch installed outside at Arctic?)
In case of abnormal temperature readed - assume sensor_read_error
hasppens.

Bug: NA (internal)


sholeksandr:
RERUN

felixradensky (approved):

PR #1522, opened 2025-01-02
hw-mgmt: thermal: Add handler for miniaml i2c bus block

sholeksandr:
• If ASIC temperature is reading as zero 3 times sequential
(during 9 secs) – don’ trust to mlxsw_minimal readouts and run emergency
flow.
• Emergency flow should enforce fan to full speed forever until
system reboot or hw-mamagement restart.
• When it happens, hw-management-tc will set to '1' file
/var/run/hw-management/config/thermal_enforced_full_spped

OS can monitor thermal_enforced_full_spped and issue tech-support dump
for future debug

Bug: 4203784, 4203785, 4219113


sholeksandr:
RERUN

sholeksandr:
RERUN

PR #1523, opened 2025-01-03
hw-mgmt: config: Fix N5XXX reset count

sholeksandr:
Fix reset count validation after removing several reset flags in
platform driver

Bug: 4237150


acoifmannvidia:
RERUN

PR #1525, opened 2025-01-06
hw-mgmgt: debug: Add debug dump handler for PWM enforce event

sholeksandr:
In case of PWM enforce event hApens (because of TC issued it) debug dump
handler will be called.

Flag:
/var/run/hw-management/config/thermal_enforced_full_spped
will set in case of FAN full speed set in abnormal situations, like
ASIC not accesible.

When it happens, hw-management will call /usr/bin/hw-managemrnt-user-dump, which is user callback. OS on init can create symbolic link:
/usr/bin/hw-managemrnt-user-dump -> /usr/cumulus/bin/cl-support.
OS can set their callback, if needed or not set at all.

Bug: 4221324, 4221324


sholeksandr:
RERUN

acoifmannvidia (approved):

PR #1531, opened 2025-01-07
V.7.0040.3000 Q3450

sholeksandr:
- hw-mgmt: thermal: add PCH thermal sensor support
- hw-mgmt: thermal: Add thermal data for QM3450


felixradensky (commented):
We don't have PSUs on Taipan, so all PSU entries look redundant.

felixradensky (approved):

PR #1541, opened 2025-01-12
V.7.0040.3000 BR

sholeksandr:
- hw-mgmt: scripts: Add ASIC state check on temperature reading
- hw-mgmt: script: Fix crash in the sync script
- hw-mgmt: script: fix update asic temp in indepenrent mode

### DIFF
--- a/usr/usr/bin/hw_management_sync.py
+++ b/usr/usr/bin/hw_management_sync.py
@@ -89,18 +89,12 @@ atttrib_list = {
          "arg": [],
          "poll": 1, "ts": 0},
 
-        {"fin": "/sys/module/sx_core/asic0/temperature/input",
-         "fn": "asic_temp_populate",
-         "arg" : ["asic"],
-         "poll": 3, "ts": 0},
-        {"fin": "/sys/module/sx_core/asic0/temperature/input",
-         "fn": "asic_temp_populate",
-         "arg" : ["asic1"],
-         "poll": 3, "ts": 0},
-        {"fin": "/sys/module/sx_core/asic1/temperature/input",
-         "fn": "asic_temp_populate",
-         "arg" : ["asic2"],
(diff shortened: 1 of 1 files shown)


sholeksandr:
RERUN

usr/usr/bin/hw_management_sync.py
```diff
@@ -542,18 +542,26 @@ def asic_temp_populate(arg_list, arg):
...
-        temp_norm = ""
+        with open(f_asic_reay, 'r') as f:
+            asic_ready = int(f.read().strip())
+    except (FileNotFoundError, ValueError):
+        asic_ready = 1
+
+    if asic_ready:
+        try:
+            val = sdk_temp2degree(int(arg))
+            temp_norm = "75000\n"
```
ciju-nvidia: Are these values same for both QTM  / SPC asics? 
If not, please consider not to hard code it.

sholeksandr: Yes. This values was also hardcoded in minimal.

sholeksandr:
RERUN

usr/usr/bin/hw_management_sync.py
sholeksandr: #define MLXSW_THERMAL_ASIC_TEMP_NORM	75000	/* 75C */
#define MLXSW_THERMAL_ASIC_TEMP_HIGH	85000	/* 85C */
#define MLXSW_THERMAL_ASIC_TEMP_HOT	105000	/* 105C */

ciju-nvidia (approved):

felixradensky (commented):
The agreement with SDK team is that SDK will report ISSU state via is ENODEV error on module sysfs file reading. So this patch should be reworked

sholeksandr:
RERUN

sholeksandr:
RERUN

PR #1545, opened 2025-01-14
hw-mgmt: script: Fix CPU VR Vcore out of range problem.

mshych:
There is a problem in VCORE output of VR on Comex board. This output depends
on CPU frequency and according to MPS vendor theoretically can be 0.
Thus reported low critical should be 0 too.
Currently problem is reported just for CFL CPU comex VR.

Bug: 4184688


mshych:
RERUN

mshych:
Just realized that 3000 branch is freezed

PR #1546, opened 2025-01-14
V.7.0040.2000 BR

mshych:
- hw-mgmt: script: Fix CPU VR Vcore out of range problem.
- Update changelog V.7.0040.2106

### DIFF
--- a/usr/usr/bin/hw-management-chassis-events.sh
+++ b/usr/usr/bin/hw-management-chassis-events.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Copyright (c) 2018 - 2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright (c) 2018 - 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions are met:
@@ -34,6 +34,7 @@
 source hw-management-helpers.sh
 board_type=$(< $board_type_file)
 sku=$(< $sku_file)
+cpu_type=$(<"$config_path"/cpu_type)
 
 LED_STATE=/usr/bin/hw-management-led-state-conversion.sh
 i2c_bus_def_off_eeprom_cartridge=7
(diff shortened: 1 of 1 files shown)


mshych:
RERUN

felixradensky (approved):

usr/usr/bin/hw-management-chassis-events.sh
```diff
@@ -920,6 +922,18 @@ if [ "$1" == "add" ]; then
 						check_n_link "$3""$4"/in"$sensor_id"_max $environment_path/"$prefix"_in"$i"_crit
 					fi
 					if [ -f "$3""$4"/in"$sensor_id"_lcrit ]; then
+						# There is a problem in VCORE output of VR on Comex board. This output depends
```
ciju-nvidia, line 925: Shouldn't we require changes in sensors.conf also? Sensors command reads from the hwmon interface and it will show the issue.

mshych: It's not required as 0 is written to driver output and sensors read from there.
E.g.
root@r-moose-03:/var/run/hw-management/environment# sensors -c /etc/hw-management-sensors/sn5600_sensors.conf| grep PMIC-12
PMIC-12 PSU 13V5 Rail (vin):           13.50 V  (crit max = +15.00 V)
PMIC-12 COMEX VCORE (out1):             1.25 V  (crit min =  +**0.00** V, crit max =  +1.58 V)

mshych:
RERUN

usr/usr/bin/hw-management-chassis-events.sh
ciju-nvidia: Usually the hwmon attributes are read only
root@r-moose-03:/sys/bus/i2c/devices/39-006b/hwmon/hwmon29# ls -l in2_lcrit
-r--r--r-- 1 root root 4096 Jan 15 03:27 in2_lcrit

Now with the 'chmod 0644' permission change, it becomes
root@r-moose-03:/sys/bus/i2c/devices/39-006b/hwmon/hwmon29# ls -l in3_lcrit
-rw-r--r-- 1 root root 4096 Jan 15 07:58 in3_lcrit

This leaves an issue that user can overwrite the value. Once if it happens, 'sensors -c' will not be able to apply it to 0.

mshych: Users with root permissions can perform everything. A report of wrong low critical is not the worst thing.

mshych:
RERUN

PR #1548, opened 2025-01-15
V.7.0040.1800 BR

mshych:
- hw-mgmt: kernel & script: add reset cause for N51xx platforms.
- Update changelog V.7.0040.1832


mshych:
RERUN

PR #1550, opened 2025-01-15
hw-mgmt: script: Fix issue in redfish client get pass

sholeksandr:
Fix race condition between hw-mgmt redfish api code and nvos bmc code,
where both try to remove this nvos_const.bin file without guard

Bug: 4241640


sholeksandr:
RERUN

PR #1554, opened 2025-01-16
hw-mgmt: scripts: Fix creation and removal of psuN_i2c_bus files

felixradensky:
These files are created when PSU cables are plugged in and removed when PSU cable are plugged out. Instead, they should be created and removed when PSUs are plugged in and plugged out respectively. Otherwise there is no way to distingush between dummy PSU and real PSU that was disconnected from power.

Bug: 4176785

### DIFF
--- a/usr/usr/bin/hw-management-chassis-events.sh
+++ b/usr/usr/bin/hw-management-chassis-events.sh
@@ -652,6 +652,31 @@ function handle_hotplug_event()
 			fi
 		fi
 		;;
+	psu*)
+		psu_name="$attribute"
+		# SN5600, SN5400 systems have PSU2 with I2C address 0x5a. In udev rules 0x5a corresponds to psu4.
+		if [[ ( $sku == "HI144" || $sku == "HI147" ) && "$2" == "psu4" ]]; then
+			psu_name="psu2"
+		fi
+		find_i2c_bus
+		i2c_comex_mon_bus_default=$(< $i2c_comex_mon_bus_default_file)
+		comex_bus=$((i2c_comex_mon_bus_default+i2c_bus_offset))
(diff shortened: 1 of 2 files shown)


sholeksandr (dismissed):

usr/usr/bin/hw-management-chassis-events.sh
```diff
@@ -652,6 +652,31 @@ function handle_hotplug_event()
...
 		;;
+	psu*)
+		psu_name="$attribute"
+		# SN5600, SN5400 systems have PSU2 with I2C address 0x5a. In udev rules 0x5a corresponds to psu4.
+		if [[ ( $sku == "HI144" || $sku == "HI147" ) && "$2" == "psu4" ]]; then
+			psu_name="psu2"
+		fi
+		find_i2c_bus
+		i2c_comex_mon_bus_default=$(< $i2c_comex_mon_bus_default_file)
+		comex_bus=$((i2c_comex_mon_bus_default+i2c_bus_offset))
```
mshych, line 663: I just wonder if we can receive such event with comex_i2c_bus.
Maybe this check is not required.

felixradensky:
RERUN

usr/usr/bin/hw-management-chassis-events.sh
felixradensky: This code was copied from hw-management-thermal-events.sh It was added by Vadim in this commit: https://github.com/Mellanox/hw-mgmt/commit/be275d91e67d4eaf17a1899e9a47f289aad0318e. There are many places where comex_i2c_bus is checked

felixradensky:
RERUN

felixradensky:
RERUN

acoifmannvidia (approved):

felixradensky:
RERUN

PR #1563, opened 2025-01-22
hw-mgmt: init: Blacklist cdc_subset network driver

felixradensky:
Initialization of usb network interface on BMC enabled systems occasionally results in the following error messages:

cdc_subset: probe of 1-2.1:1.0 failed with error -22
cdc_ether: probe of 1-2.1:1.0 failed with error -16

Since both cdc_subset and cdc_ether drivers can handle usb network interface between host CPU and BMC, blacklist the cdc_subset driver to make sure that only cdc_ether driver is loaded. This fixes the error messages above.

Bug: 4260024


sholeksandr (approved):

felixradensky:
RERUN

PR #1569, opened 2025-01-27
hw-mgmt: patches: 5.10/6.1: Add kernel new kernel version support

sholeksandr:
Add kernel 5.10.226/6.1.123 support


sholeksandr:
RERUN

PR #1573, opened 2025-01-28
hw-mgmt: sensor: Fix for unexpected negative vout_min on lm5066i

sholeksandr:
By default in lm5066i VOUT_UV_WARN_LIMIT registrer == 0. Because of issue in
driver calculation formula it trasnsforms to -0.11V instead of 0, which is
not reasonable.
It can be fixed by writing 0 to *_in2_min sysfs attribute.

Bug: 4264364


sholeksandr:
RERUN

acoifmannvidia (approved):

PR #1582, opened 2025-02-03
hw-mgmt: thermal: TC fix for sensor min/max limits calculation

sholeksandr:
For systems that use sensor value scale factor != 1000 this
value should be applied to min/max sensor threshold.
This commit fix bug with min/max calculation for sensors with
value divider differents than default (1000)

Bug: 4284991


sholeksandr:
RERUN

PR #1584, opened 2025-02-03
hw-mgmt: rules: Fix initialization of usb network interface

felixradensky:
Udev rule to initialize usb network interface can run before configuration file /etc/network/interfaces was created or before the configuration data for usb network interface was added to this file. This can result in the following error messages:

(udev-worker): usb0: Process ‘/usr/sbin/ifup usb0’ failed with exit code 89.

Add checks to avoid such errors from showing in the logs.

Bug: 4260031


felixradensky:
RERUN

sholeksandr (approved):

vadimp-nvidia (approved):

PR #1589, opened 2025-02-09
hw-mgmt: rules: Explicitly set the name of USB network interface

felixradensky:
In kernel version 6.1.123 the name of USB network interface has changed from usb0 to eth2. This breaks IP address assignement of this interface as our network configuration files assume usb0 as interface name. Fix udev rule to explicitcly set the name of this interface as usb0.


felixradensky:
RERUN

acoifmannvidia (approved):

PR #1592, opened 2025-02-10
hw-mgmt: sensors: Remove MDIO PHY from sensors config

vadimp-nvidia:
Set to 'ignore' PHY MDIO on SN4280 and SN5640, since reading ofthis device is not stable.
And it is not considered for thermal monitoring.


vadimp-nvidia:
RERUN

felixradensky (approved):

PR #1595, opened 2025-02-12
hw-mgmt: sensors: Remove MDIO PHY from sensors config

felixradensky:
Set to 'ignore' PHY MDIO on SN4280 and SN5640, since reading of this device is not stable.
And it is not considered for thermal monitoring.

Bug: 4037463


felixradensky:
RERUN

sholeksandr (approved):

vadimp-nvidia (approved):

PR #1596, opened 2025-02-12
Added kernel patches for interrupt storm protection.

bspguy:
mlxreg-hotplug: Add support for handling interrupt storm. Updated patch status table.


vadimp-nvidia:
RERUN

PR #1608, opened 2025-02-17
hw-mgmt: script: Optimize sync script module update

sholeksandr:
Optimize sync script module update.

1. Change in system configuration. Define total module count instead
   of per module definition.
2. Check SDK SW/FW reading mode by reading just one module config.
   No need to check each module config because modules can't be initialized
   in the "mixed" mode.


felixradensky (approved):

sholeksandr:
RERUN

PR #1610, opened 2025-02-18
V.7.0040.2000 BR

sholeksandr:
- hw-mgmgt: thermal: TC add more debug printouts
- Add support of lcrit/hcrit to asic/amb sensor
- hw-mgmt: thermal: TC fix blacklist not working issue
- hw-mgmt: thermal: TC fix for sensor min/max limits calculation


sholeksandr:
RERUN

PR #1611, opened 2025-02-18
hw-mgmt: script: fix issue sync service does not create module sysfs

sholeksandr:
Fix issue: sync service does not create module sysfs

Bug: 4314895


sholeksandr:
RERUN

felixradensky (approved):

PR #1631, opened 2025-03-04
hw-mgmgt: thermal: Fix PSU_FAN speed control for FW controlled PSU

sholeksandr:
If PSU FAN should be controlled by its FW we shouldn't set default PSU
FAN speed on init. PSU FAN default spped will be set only for systems
with TC controllable PSU FANs

Bug: N/A


sholeksandr:
RERUN

acoifmannvidia:
tested?

felixradensky (commented):
It test description please change spped/speed

felixradensky (approved):

sholeksandr:
RERUN

PR #1632, opened 2025-03-05
hw-mggmt: thermal: Fix thermal amb sensor link for multi asic systems

sholeksandr:
Fix thermal sensor link create for multi asic systems. For multi ASIC
systems hw-mgmt should check if i2c device doesn't belong to any ASIC
bus.

Bug: 4327177


sholeksandr:
RERUN

PR #1636, opened 2025-03-07
hw-mgmt: scripts: Enable Juliet platform emulation

ciju-nvidia:
Bug: N/A

This patch set adds the support for
 - Enabling platform emulation for Juliet
 - Sensor attributes for which the sensor emulation driver is not available.
 - Workaround for sodimm sensor availability in AMD based virtual platforms


ciju-nvidia:
RERUN

acoifmannvidia (approved):

PR #1649, opened 2025-03-19
hw-mgmt: scripts: Enable Juliet platform emulation

akomisar-nv:
Bug: 4363327

This patch updates values and adds missing values to voltage sensor data on Juliet Qemu emulation.


usr/etc/hw-management-virtual/HI166/environment
```diff
@@ -30,3 +30,71 @@ pdb_hotswap1_in2_lcrit 0
 pdb_hotswap1_in2_min 0
```
ciju-nvidia, line 30: Please update the commit heading. Say something like given below. Also move the bug id after the description.

====
hw-mgmt: scripts: Update platform emulation sensor values for Juliet

This patch updates values and adds missing values to voltage sensor data
on Juliet Qemu emulation.

Bugs: 4363327

Signed-off-by: Aaron Komisar <akomisar@nvidia.com>
====

PR #1657, opened 2025-03-25
hw-mgmt: patches: Improve dpu interrupt handling

ciju-nvidia:
DPU auxiliary powering can cause interrupt flooding because DPU interrupt handlers are not configured yet, while middle interrupt aggregation register is unmasked by default during initialization. Thus, interrupts are getting through, while handlers are still not fully initialized. Do not unmask aggregation interrupt register at initialization for all DPUs. Instead do it per DPU, when its initialization is done.

This patch also adds the change to clear the DPU event registers.

This commit also does the following:
 - There are patches on mlx-platform.c which are not taken by certain NOSs. This commit rebases the patches and update the patch_status table to avoid build failures.
- Patch rebase to avoid conflicts in mlx-platform.c

Bugs: 4338675


ciju-nvidia:
RERUN

ciju-nvidia:
RERUN

ciju-nvidia:
RERUN

ciju-nvidia:
RERUN

ciju-nvidia:
RERUN

felixradensky (approved):

PR #1658, opened 2025-03-25
hw-mgmt: scripts: Add detection of dummy PSUs on Q3200/Q3400 systems

felixradensky:
Dummy PSUs may be installed instead of real PSUs to ensure the proper chassis air flow on systems where some real PSUs are missing. Such PSUs have presence bit but do not show up on I2C bus. To differentiate them from real PSUs that are not connected to power add the code to detect them during boot and hotplug and create psuN_is_dummy attributes.

Bug: 4343164

### DIFF
--- a/usr/usr/bin/hw-management-chassis-events.sh
+++ b/usr/usr/bin/hw-management-chassis-events.sh
@@ -629,6 +629,44 @@ function handle_hotplug_dpu_event()
     fi
 }
 
+function handle_hotplug_psu_event()
+{
+	local psu_name=$1
+	local event=$2
+	local psu_num
+	local psu_i2c_bus
+	local psu_i2c_addr
+	local dummy_psus_supported=$(< ${config_path}/dummy_psus_supported)
+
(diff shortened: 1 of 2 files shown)


usr/usr/bin/hw-management-chassis-events.sh
```diff
@@ -629,6 +629,44 @@ function handle_hotplug_dpu_event()
...
+	local event=$2
+	local psu_num
+	local psu_i2c_bus
+	local psu_i2c_addr
+	local dummy_psu_supported=$(< ${config_path}/dummy_psu_supported)
+
+	if [ ${dummy_psu_supported} -eq 1 ]; then
+		case ${sku} in
+		HI157)
+			psu_i2c_bus=(4 4 4 4)
```
vadimp-nvidia, line 644: We already have this info in psu_i2c_map. Maybe we can have it on one single location? Maybe save it config folder and read from there?
However also this way is OK.

vadimp-nvidia (approved):

felixradensky:
RERUN

PR #1661, opened 2025-03-26
hw-mgmt: Increase fast sysfs T.O

bspguy:
Increase fast sysfs T.O for supporting lower end systems. Exit with warning in case of T.O


felixradensky (approved):

felixradensky:
RERUN

PR #1671, opened 2025-04-06
hw-mgmt: sysvinit files

bspguy:
Added lsb block header for older systems with sysvinit support, in sysfs monitor init files.


sholeksandr (approved):

felixradensky:
RERUN

PR #1694, opened 2025-04-14
hw-mgmgt: scripts: Fix ASIC tempreature sync

sholeksandr:
Fix ASIC tempreature sync

Bug: 4238116


sholeksandr:
RERUN

PR #1706, opened 2025-04-17
hw-management: fast sysfs monitor

bspguy:
Add protection in case race condition between
hw-management service and fast-sysfs service.

Bug: 4406920


felixradensky:
RERUN

PR #1707, opened 2025-04-17
hw-management: sysfs systemd services

bspguy:
Remove TimeoutStopSec=1.
Timeout will be handled by systemd default.

Bug: 4406833


felixradensky (approved):

felixradensky:
RERUN

PR #1717, opened 2025-04-27
hw-mgmt: scripts: Fix dummy PSU detection during PSU hotplug events

felixradensky:
When real PSU is inserted, it takes several seconds for relevant I2C device to be created. Dummy PSUs have no I2C device associated with them. The code that checks if inserted PSU is dummy does not take the aforementioned delay into account and can erroneously mark real PSU as dummy. Fix that by adding retries and delays to dummy PSU detection code.

Bug: 4176785


felixradensky:
RERUN

vadimp-nvidia (approved):

PR #1719, opened 2025-04-28
hw-mgmgt: scripts: Improve debug dump

sholeksandr:
Add CPLD registers dump "cpld_reg_direct_dump" filem created with
the direct reading "iorw" cmd.

Bug: 4421390


sholeksandr:
RERUN

PR #1721, opened 2025-04-29
hw-mgmt: kernel: config: Define CONFIG_REGMAP_I2C as built-in

felixradensky:
Other kernel config options that depend on CONFIG_REGMAP_I2C (e.g. CONFIG_I2C_MUX_PCA954x) are defined as built-in, so CONFIG_REGMAP_I2C should be also defined as built-in.

Bug: 4424482


felixradensky:
RERUN

mshych (approved):

felixradensky:
RERUN

PR #1726, opened 2025-05-01
hw-mgmt: service: Fix deadlock for hw-gmgt <-> TC sevice start on SIMX

sholeksandr:
On SIMX, hw-mgmt tries to start the TC service during initialization. However, the TC service has a strong dependency:
Requires=hw-management.service. This means the TC service cannot start before hw-mgmt starts. But hw-mgmt attempts to start TC earlier. In this case, neither hw-mgmt nor TC will complete initialization.

This commit resolves this deadlock by postponing the start of the TC service.

Bug: 4427090

### DIFF
--- a/usr/usr/bin/hw-management-helpers.sh
+++ b/usr/usr/bin/hw-management-helpers.sh
@@ -303,6 +303,18 @@ check_simx()
 	fi
 }
 
+# This function checks if ThermalControl supports current platform
+check_tc_support()
+{
+    sys_sku=$(<$sku_file)
+    if ([ "$sys_sku" = "HI166" ] ||
+        [ "$sys_sku" = "HI167" ]); then
+        return 1
+    else
+        return 0
+    fi
+}
+
 # This function create or cleans sysfs monitor helper files.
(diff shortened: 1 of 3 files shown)


sholeksandr:
RERUN

sholeksandr:
RERUN

usr/usr/bin/hw-management-helpers.sh
```diff
@@ -303,6 +303,18 @@ check_simx()
 	fi
 }
 
+# This function checks if ThermalControl supports current platform
+check_tc_support()
+{
+    sys_sku=$(<$sku_file)
+    if ([ "$sys_sku" = "HI166" ] ||
+        [ "$sys_sku" = "HI167" ]); then
```
akomisar-nv, line 311: The original code which this patch removes from the start-post script disables TC for SKU's HI176 and HI177. This function reports TC is not supported for SKU HI166 and HI167. Is this intentional?

sholeksandr: Yes. It is intentional

felixradensky (approved):

PR #1733, opened 2025-05-06
hw-mgmt: service: Define the type sysfs monitor services as oneshot

felixradensky:
Without oneshot definition Sonic service monitor shows the service status as "Down"

Bug: 4389689


felixradensky:
RERUN

sholeksandr:
RERUN

sholeksandr (approved):

PR #1744, opened 2025-05-11
hw-mgmt: service: Fix dependencies of hw-mgmt fast sysfs monitor service

felixradensky:
The service depends of directory structure created by hw-mgmt service and should start running after hw-mgmt service.

Bug: 4439383


felixradensky:
RERUN

PR #1745, opened 2025-05-12
hw-mgmt: service: Fix dependency of hw-mgmt-fast-sysfs-monitor service

bspguy:
Since this service is a 'oneshot' type service,
Remove dependency with hw-management main service. Also add protection in case directory structure
is not ready.

Bug: 4439383


felixradensky (approved):

felixradensky:
RERUN

felixradensky:
RERUN

PR #1755, opened 2025-05-15
hw-mgmt: scripts: Add attribute with SED PBA version.

mshych:
New SED PBA, versions 83.01.0008 and up, reports version through EFI variable.
If applicable hw-management reads this ERFI variable and report to NOS.


mshych:
RERUN

felixradensky (approved):

PR #1759, opened 2025-05-16
hw-mgmgt: thermal: TC add support module with TEC cooling

sholeksandr:
TC add support module with TEC cooling.


sholeksandr:
RERUN

PR #1772, opened 2025-05-20
hw-mgmgt: thermal: TC: Fix default smoothing level for sensor input

sholeksandr:
TC has an option to smooth sensor input values (to suppress fast PWM
jumps).
By default, "No smoothing" should be used for sensor input values.
There was an error in the smoothing formula which unintentionally
enabled the minimum smoothing level for all sensors.

This fix sets the "No smoothing" level as the default for all sensor
inputs.
This is not a critical bug.

Bug: 4452512


sholeksandr:
RERUN

acoifmannvidia (approved):

PR #1782, opened 2025-05-21
V.7.0040.4000 BR sholeksandr1295

sholeksandr:
- hw-mgmgt: thermal: TC: Fix default input val smoothing level
- hw-mgmt: thermal: TC fix for asic sensor read error handling


sholeksandr:
RERUN

PR #1784, opened 2025-05-21
hw-mgmgt: thermal: TC: Fix default input val smoothing level

sholeksandr:
- hw-mgmgt: thermal: TC: Fix default input val smoothing level
- hw-mgmt: thermal: TC fix for asic sensor read error handling
- hw-mgmgt: thermal: TC: Fix default input val smoothing level


sholeksandr:
RERUN

PR #1786, opened 2025-05-27
hw-mgmgt: script: Fix asic temp sync

sholeksandr:
Fix asic temp sync for systems with minimal disabled.
ASIC temperature sync should be updated independent of SDK mode.

Bug: 4463009


sholeksandr:
RERUN

PR #1788, opened 2025-05-27
hw-mgmt: thermal: Fix TC service disable on system start

sholeksandr:
By mistake we had "reversed logic" for disabling ThermalControl
on systems without TC support.

Bug: 4466832


sholeksandr:
RERUN

PR #1792, opened 2025-05-28
hw-mgmt: thermal: TC add flag to disable tec support

sholeksandr:
Added a flag to the TC to disable TEC feature support by default. It
should be removed after the qualification tests.


sholeksandr:
RERUN

PR #1802, opened 2025-06-01
hw-mgmt: thermal: Fix code syntax issue in TC

sholeksandr:
Fix code syntax issue in TC. Missing ':' symbol

Bug: 4474282


sholeksandr:
RERUN

vadimp-nvidia (approved):

PR #1811, opened 2025-06-02
hw-mgmgt: scripts: Add cpld registers dump on system start

sholeksandr:
Add cpld registers dump on system start. Dump will be
saved in the system 'syslog' logfile.

FR: 4458180


sholeksandr:
RERUN

PR #1812, opened 2025-06-02
hw-mgmgt: scripts: Add cpld registers dump to cyclic rotating log file (2 commits)

sholeksandr:
- hw-mgmgt: scripts: Add cpld registers dump on system start
- hw-mgmgt: scripts: Add cpld registers dump to cyclic rotating log file


sholeksandr:
RERUN

PR #1825, opened 2025-06-08
hw-mgmgt: thermal: Fix FAN min/max speed settings

sholeksandr:
For system FANs we have attributes fanX_min/fanX_max which show FAN
max/min speed. But for systems with 2 FANs in a drawer, speeds are differ
for front/rear fan.

The current limits do not take this into account, and the same values
are used for both rear and front fans. This causes issues when the
actual maximum fan speed and the fanX_max attribute differ by more than
10%.

To address this, separate configurations should be used for the front
and rear fans.

This commit adds separate min/max fan limits for SN2700 and SN3700
systems. The actual fan min/max speeds can vary by ±10% from the
configured limits.

SN2700
front min - 21000
front max - 6300
rear min - 18000
rer max - 5400

SN3700
front min - 23000
front max - 5400
rear min - 20500
rer max - 4800

Bug: 4468363


acoifmannvidia:
hi @sholeksandr , pls elaborate how tested

sholeksandr:
Tested on different (5) systems. In manual mode - for SN2700/3700 check that odd/even fans has different FAN speed limits according to predefined configuration .

For other systems checked that limits where not changed compared to version without the fix.

Checked FAN limits after complete system reboot and after hw-mgmgt service restart

sholeksandr:
RERUN

sholeksandr:
RERUN

PR #1838, opened 2025-06-12
hw-mgmt: scripts: Fix module temperature sync conflict in SW mode

sholeksandr:
If the module is in SW mode, hw-mgmt shouldn't update the
./thermal/moduleX_* attributes.

In our algorithm, if a module is not present, we fill its attributes with zeros.
In SW mode, the "present" status is "0", which caused a conflict with a
similar OS process when hw-mgmt overwrote the module attributes with
zeros.
If affect thermal algo because it try to set PWM based on wrong inputs
(module temperature is 0).

In SW mode, hw-mgmt shouldn't even check the module status and should
give priority to the OS for updating the attributes.

This commit adds a check for the module's "mode" first, and if it's in SW mode,
skips the update. We don't even need to read the "present" status in this case

Bug: 4494119

Unittests incuded.

### DIFF
--- a/unittest/hw_mgmgt_sync/module_populate/README_tests.md
+++ b/unittest/hw_mgmgt_sync/module_populate/README_tests.md
@@ -1,0 +1,120 @@
+# Unit Tests for module_temp_populate Function
+
+This directory contains comprehensive unit tests for the `module_temp_populate` function from `hw_management_sync.py`.
+
+## Features
+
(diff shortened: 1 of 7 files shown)


acoifmannvidia (approved):

sholeksandr:
RERUN

sholeksandr:
RERUN

PR #1841, opened 2025-06-12
hw-mgmt: scripts: Adjust the check of usb0 interface configuration

felixradensky:
Currently the validity checks for usb0 interface configuration assime that it is defined as "allow-hotplug usb0" in cofiguration file /etc/network/interfaces. Some NOSes may define it as "auto usb0", and validity checks will fail. This can result in usb0 remaining unconfigured on host side after BMC reboot. Adjust the checks to allow "auto usb0".

Bug: 4491046


acoifmannvidia (approved):

sholeksandr:
RERUN

PR #1862, opened 2025-06-17
hw-mgmt: scripts: Create PSU EEPROM symlinks only if actual files exist

felixradensky:
In case hw-mgmt fails to initialize PSU EEPROM driver, the eeprom file under /sys will not be created. In that situation PSU EEPROM symlink under hw-mgmt tree should not be created to avoid generating dangling links.

Bug: 4478004

### DIFF
--- a/usr/usr/bin/hw-management-thermal-events.sh
+++ b/usr/usr/bin/hw-management-thermal-events.sh
@@ -986,7 +986,11 @@ if [ "$1" == "add" ]; then
 				psu_eeprom_type=$(< "$config_path"/psu_eeprom_type)
 			fi
 			echo "$psu_eeprom_type" 0x"$psu_eeprom_addr" > /sys/class/i2c-dev/i2c-"$bus"/device/new_device
-			ln -sf "$eeprom_file" "$eeprom_path"/"$eeprom_name" 2>/dev/null
+			for ((i=0; i<2; i++)); do
+				[ -f "$eeprom_file" ] && break
+				sleep 0.5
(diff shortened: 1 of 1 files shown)


felixradensky:
RERUN

usr/usr/bin/hw-management-thermal-events.sh
```diff
@@ -986,7 +986,11 @@ if [ "$1" == "add" ]; then
 				psu_eeprom_type=$(< "$config_path"/psu_eeprom_type)
 			fi
 			echo "$psu_eeprom_type" 0x"$psu_eeprom_addr" > /sys/class/i2c-dev/i2c-"$bus"/device/new_device
-			ln -sf "$eeprom_file" "$eeprom_path"/"$eeprom_name" 2>/dev/null
+			for ((i=0; i<5; i++)); do
+				[ -f "$eeprom_file" ] && break
+				sleep 1
```
ciju-nvidia, line 991: If the eeprom is not available (when driver fails to instantiate), this will add 5sec delay. This may be an issue for sysfs monitoring. Please accomodate that when committing to master branch.

sholeksandr: I think wait 2 times with delay 0.5 sec will be enough. It will add maximum 1 sec delay

felixradensky: Fixed, please approve

felixradensky:
RERUN

PR #1881, opened 2025-06-24
Rosalind bu simx

akomisar-nv:
Add simx support for Rosalind.


felixradensky (commented):
1. We don't have SODIMM temp sensors on Rosalind, since we don't have SODIMM
2. Please rename hwmgmt.tar to hwmgmt_HI180.tgz
3. Please rename Rosalind to HI180 in hw-management-ready.sh
4. We should also have votmon7-voltmon16 in environment

usr/usr/bin/hw-management.sh
```diff
@@ -3729,6 +3729,12 @@ case $ACTION in
 			log_err "hw-management is already started"
 			exit 1
 		fi
+		#TEMPORARY hw-management mockup values for Rosalind in simx
```
ciju-nvidia, line 3732: Remove this code block.  
If it's for UI tree, please refer to the existing code for extracting UI archives for other QM platforms (Eg: Crocodile, Gorilla, Juliet, etc.).

akomisar-nv: This is not for the UI tree, it is for the entire /var/run/hw-management.
Currently there is no support for native i2c in qemu. Until we handle this issue, i2c peripherals shall not be supported in simx and therefore we need to provide static files.

usr/etc/hw-management-virtual/hwmgmt.tar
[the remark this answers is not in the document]
akomisar-nv: This tar file is not for the ui tree. It is for the entire /var/run/hw-management.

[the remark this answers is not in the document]
akomisar-nv: File renamed to hwmgmt_HI180.tgz.

usr/usr/bin/hw-management.sh
akomisar-nv: File renamed to hwmgmt_HI180.tgz.

PR #1882, opened 2025-06-24
hw-mgmt: scripts: Fix ASIC temperature reading issue

sholeksandr:
If we try to read asic temperature before SDK fully initialized
It can may result in error message in kernel:

sxd_kernel: [error] Command path in DPT for device 1 is not valid.
Aborting command SX_CMD_ACCESS_REG (register ID 0x900f)
sxd_kernel: [error] Register access failed (reg_id=0x900f, err=-22 [from CMD-IFC
status])

Fix: hw-management should use sysfs only after these events:

KOBJ_ADD or KOBJ_ONLINE from sx_core driver
and should not use sysfs after these events
KOBJ_REMOVE or KOBJ_OFFLINE from sx_core driver

Bug: 4497061

### DIFF
--- a/usr/usr/bin/hw-management-thermal-events.sh
+++ b/usr/usr/bin/hw-management-thermal-events.sh
@@ -275,6 +275,33 @@ function get_psu_eeprom_type()
 	echo $eeprom_type
 }
 
+# Set ASIC ready state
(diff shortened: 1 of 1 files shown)


sholeksandr:
RERUN

PR #1883, opened 2025-06-24
hw-mgmt: scripts: Fix ASIC temperature reading issue

sholeksandr:
If we try to read asic temperature before SDK fully initialized
It can may result in error message in kernel:

sxd_kernel: [error] Command path in DPT for device 1 is not valid.
Aborting command SX_CMD_ACCESS_REG (register ID 0x900f)
sxd_kernel: [error] Register access failed (reg_id=0x900f, err=-22 [from CMD-IFC
status])

Fix: hw-management should use sysfs only after these events:

KOBJ_ADD or KOBJ_ONLINE from sx_core driver
and should not use sysfs after these events
KOBJ_REMOVE or KOBJ_OFFLINE from sx_core driver

Add asic{X}_ready flag which can be used to distinguish if ASIC initialized

Bug: 4497061

### DIFF
--- a/usr/usr/bin/hw-management-thermal-events.sh
+++ b/usr/usr/bin/hw-management-thermal-events.sh
@@ -275,6 +275,33 @@ function get_psu_eeprom_type()
 	echo $eeprom_type
 }
 
+# Set ASIC ready state
+# $1 - sysfs path of PCI device
+# $2 - ASIC ready state (0 or 1)
+# return none
+function set_asic_ready()
+{
+	sysfs_picdev_path=$1
+	state=$2
+	[ -f "$config_path/asic_num" ] && asic_num=$(< $config_path/asic_num)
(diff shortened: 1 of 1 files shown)


usr/usr/bin/hw-management-thermal-events.sh
```diff
@@ -1142,6 +1169,7 @@ if [ "$1" == "add" ]; then
 			modprobe mlxsw_minimal
 		fi
 		/usr/bin/hw-management.sh chipup 0 "$4/$5"
+		set_asic_ready "$4/$5" 1
```
ciju-nvidia, line 1172: Please ignore if this point is already tried:
The issue was reported on moose-dgx-05 switch with some cables connected. I think, it's better to recreate the issue once and see if the proposed approach really fixes the issue.

PR #1894, opened 2025-06-26
hw-mgmt: scripts: Fix ASIC temperature reading issue (3 commits)

sholeksandr:
- hw-mgmt: kernel: Fix CONFIG flags used for 'upstream' builds
- hw-mgmt: thermal: Revert FAN max/min speed attributes
- hw-mgmt: scripts: Fix ASIC temperature reading issue


sholeksandr:
RERUN

PR #1896, opened 2025-06-26
hw-mgmt: scripts: Fix voltage threshold calculation for Delta 1100 PSU

felixradensky:
The fix for bug 3874682 has added voltage threshold calculations for Delta 1100 PSU based on the value of critical max threshold. However the reading of this value was done incorrectly, resulting in all calculated theresholds being defined as 0. This patch fixes the threshold calculation.

Bug: 4505817


felixradensky:
RERUN

sholeksandr (approved):

felixradensky:
RERUN

felixradensky:
RERUN

PR #1911, opened 2025-07-02
hw-mgmt: scripts: fix init errors of GPIOs for JTAG on AMD based syst…

felixradensky:
…ems.

CPLD update on AMD based systems is done through LPC CPLD interface. So, GPIOs for JTAG are not relevant and shouldn't be initialized.

Bug: 4512069


felixradensky:
RERUN

felixradensky:
RERUN

PR #1937, opened 2025-07-11
hw-mgmt: thermal: Fan state validation after re-insertion

ciju-nvidia:
A customer (case #4486661) was reported regarding
the fan start issues after re-insertion. After re-inserting
fans in some systems, it may fail to start rotating.
    
    A stuck fan can start rotating by either of the following actions:
      1) Pulling out and reinserting an additional fans.
      2) Lowering the PWM of all fans for a few seconds.
    
    The commit ff627 introduced FR #4521169 to Thermal
    control 2.5. This patch enables the feature for Thermal
    control 2.0.
    
    Bugs:  #4360106
    NVBug: #5317658
    

sholeksandr:
RERUN

PR #1954, opened 2025-07-17
hw-mgmt: scripts: Move shared python code to separate module

sholeksandr:
Moved logger/timer classes from TC files to python module:

./usr/bin/hw_management_lib.py


sholeksandr:
RERUN

sholeksandr:
RERUN

usr/usr/bin/hw_management_lib.py
```diff
@@ -0,0 +1,282 @@
+# Copyright (c) 2019-2024 NVIDIA CORPORATION & AFFILIATES.
```
ciju-nvidia, line 1: I think it would be better to create a directory 'lib' and package it in hw-mgmt.
'usr/usr/lib/python/hw_mgmt_tc_lib.py'

In this way, it will look neat and the same folder can be used for similar python libraries in future

sholeksandr:
RERUN

sholeksandr:
RERUN

sholeksandr:
RERUN

sholeksandr:
RERUN

sholeksandr:
RERUN

PR #1960, opened 2025-07-18
hw-mgmt: sensors: Fix module counter for SN5600 switch

sholeksandr:
Fix module counter for SN5600 switch. Switch have 65 ports/modules.

Change module count:
66 -> 65

Bug: 4540295


sholeksandr:
RERUN

PR #1969, opened 2025-07-21
hw-mgmt: scripts: Fix the number of ports for SN5400 system

felixradensky:
SN5400 system has 66 front panel ports, fix the number of ports in sync service

Bug: 4540295


felixradensky:
RERUN

sholeksandr (approved):

PR #1970, opened 2025-07-21
hw-mgmt: thermal: Fix not recover PWM after asic sensor reading error

sholeksandr:
In TC when simulating asic read error we continuously see error logs,
even after asic error recovered.
This commit fix error.

Bug: 4545207


sholeksandr:
RERUN

felixradensky (approved):

PR #1972, opened 2025-07-21
hw-mgmt: thermal: add moduleX_status which show same as exposed by the SDK

sholeksandr:
add moduleX_status which will show module status same as exposed by the
SDK: /sys/module/sx_core/asic0/moduleX/present

FR: 4458339


sholeksandr:
RERUN

PR #1977, opened 2025-07-25
hw-mgmt: thermal: Fix TC init/close flow issue

sholeksandr:
In rare cases, when stopping the TC service immediately after it starts,
it may crash. This occurs because the shutdown cleanup flow uses the
system_config variable before it has been initialized.

This fix adds proper initialization of this variable to ensure
it is ready before use.

Also added optimization for logger close on TC stop: removed the
redundant
log handler flush() call. It is unnecessary because flush() is already
called within log handler stop().

Bug: 4545880


sholeksandr:
RERUN

sholeksandr:
RERUN

sholeksandr:
RERUN

sholeksandr:
RERUN

acoifmannvidia (approved):

sholeksandr:
RERUN

PR #1978, opened 2025-07-25
hw-mgmt: service: Increase TC service stop timeout

sholeksandr:
On some low-performance systems, stopping the hw-management-tc service
took longer than 1 second.
Fix: Increased the TimeoutStopSec value in the service file from 1 to 5
seconds.

Bug: 4543074


sholeksandr:
RERUN

PR #1983, opened 2025-07-29
hw-mgmt: scripts: Update SN5640 virtual sensor attributes

ciju-nvidia:
Adding the missing sensor (for the sensors which doesn't have the emulation drivers) attributes as static values in SN5640 platform. This will be helpful for running TC.

Bugs: #4546019, #4545909


ciju-nvidia:
RERUN

PR #1984, opened 2025-07-29
hw-mgmt: thermal: TC add lcrit/hcrit validation for module sensor

sholeksandr:
TC add lcrit/hcrit check for module sensor. Set "sensor read"
error in case of module temp not in range [lcrit..hcrit] (same as
for other sensors)

Bug: 4528292

### DIFF
--- a/usr/usr/bin/hw_management_thermal_control.py
+++ b/usr/usr/bin/hw_management_thermal_control.py
@@ -1636,8 +1636,8 @@ class thermal_module_sensor(system_device):
         """
         status = True
 
-        if self.last_value == 0 and self.val_max == 0 and self.val_min == 0:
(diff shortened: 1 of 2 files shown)


acoifmannvidia (approved):
all question answered by chat:
pls have a common file with common functionality between 2.5 and 2.0
This is not a case to code refactoring.  As I mentioned, it's cherry-pick from master/4000 branch.
I don't want to modify this commit.
Separation can be done in separate PR. Not in this.
pls state how was tested
Tested In our regression. It is tested during several weeks together with 4000 branch. 
pls state if was code reviewed by cursor ( it is a must)
Code reviewed in Cursor,
pls explain root cause and how fixed on the bug itself
Ok.

PR #1995, opened 2025-08-07
hw-mgmt: udev rules: disable TPM runtime power management.

mshych:
Disable TPM runtime power management, i.e. TPM will always remain power-on.
It will allow avoiding rare cases of access to TPM when it's suspended.
The effect of always power-on the TPM on common power consumption is minimal.

Bug #4542714


mshych:
RERUN

mshych:
RERUN

mshych:
RERUN

mshych:
RERUN

PR #2003, opened 2025-08-13
hw-mgmt: script: Fix ThermalControl service disable logic

sholeksandr:
On systems that do not support thermal control, the thermal service
should be disabled. By mistake, we had the logic for thermal control
support detection inverted.

This commit fixes the TC support check logic:

Return 1 if TC is supported

Return 0 otherwise

Bug: 4567911


sholeksandr:
RERUN

acoifmannvidia (approved):

PR #2004, opened 2025-08-13
hw-mgmt: topo: Fix reset_ready not set for qm3401_LD system

sholeksandr:
For qm3401_LD we have only 17 reset causes. But defined 18.
In this case reset_ready not set because real reset_cause lower then
defined.

Fix:
change reset_attr_num attribute in ./config
18 -> 17.

Bug: 4577722


vadimp-nvidia (approved):

sholeksandr:
RERUN

PR #2015, opened 2025-08-20
Hw-mgmt: kernel 5.10 6.1: patch: Fix inconsistent name for PDB items

sholeksandr:
For DGX Systems without PSU should be used "pdb" suffix in config const name.

Change:
"mlxplat_mlxcpld_dgx_psu_items_data" -> "mlxplat_mlxcpld_dgx_pdb_items_data"

Spell fix. No functional impact.


sholeksandr:
RERUN

PR #2016, opened 2025-08-20
hw-mgmt: kernel: Add support for Q3401-RD platform

sholeksandr:
Add support for Q3401-RD platform to kernel 5.10/6.1


sholeksandr:
RERUN

PR #2032, opened 2025-09-02
hw-mgmt: scripts: Add I2C bus ownership check and reporting for VMOD0021 systems.

mshych:
Implement a check for I2C bus ownership on VMOD0021 systems to determine
if it is controlled by the CPU or BMC.
Typically, ownership should be with the CPU; if not, report an error.

Bug: 4597390

### DIFF
--- a/usr/usr/bin/hw-management.sh
+++ b/usr/usr/bin/hw-management.sh
@@ -2444,6 +2444,21 @@ smart_switch_common()
 
 n51xxld_specific()
 {
+	# Report I2C bus ownership for VMOD0021 systems at early initialization
+	bmc_to_cpu_ctrl_path=$(ls /sys/devices/platform/mlxplat/mlxreg-io/hwmon/hwmon*/bmc_to_cpu_ctrl 2>/dev/null | head -n1)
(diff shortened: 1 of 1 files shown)


mshych:
RERUN

usr/usr/bin/hw-management.sh
```diff
@@ -2444,6 +2444,21 @@ smart_switch_common()
 
 n51xxld_specific()
 {
+	# Report I2C bus ownership for VMOD0021 systems at early initialization
+	bmc_to_cpu_ctrl_path=$(ls /sys/devices/platform/mlxplat/mlxreg-io/hwmon/hwmon*/bmc_to_cpu_ctrl 2>/dev/null | head -n1)
+	if [ -f "$bmc_to_cpu_ctrl_path" ]; then
+		bus_ownership=$(< "$bmc_to_cpu_ctrl_path")
+		if [ "$bus_ownership" = "0" ]; then
```
ciju-nvidia, line 2451: Even though this situation is very less likely, what's the point in continuing further at this point? Shouldn't we be exiting?

vadimp-nvidia (commented):
I think it is fine to go with this addition fir current situation.

PR #2033, opened 2025-09-03
hw-mgmt: scripts: Add I2C bus ownership check and reporting for VMOD0021 systems.

mshych:
Implement a check for I2C bus ownership on VMOD0021 systems to determine
if it is controlled by the CPU or BMC.
Typically, ownership should be with the CPU.
If not, report an error and try ownership enforcement to CPU.

Bug: 4597390


mshych:
RERUN

mshych:
RERUN

mshych:
RERUN

PR #2051, opened 2025-09-17
hw-mgmt: scripts: Correct dependency for systemd startup

ciju-nvidia:
The hw-management-sysfs-monitor does not automatically
start when hw-mgmt is restarted. It is because the
PartOf= directive only causes stop and restart operations
on the main service to be propagated to the dependent. It
does not cause a standard "start" or "restart" to also
start the dependent if it was previously inactive.
If hw-management-sysfs-monitor is not active at the
moment hw-mgmt is restarted, it will stay inactive.

To make sure hw-management-sysfs-monitor also (re)starts
whenever hw-mgmt is started or restarted, adding
'Wants=' directive to hw-management service.

Bug: 4540628, 4556756

### DIFF
--- a/debian/Release.txt
+++ b/debian/Release.txt
@@ -1,4 +1,29 @@
 ================================================================================
+- V.7.0050.1934
+- Wed, 17 Sep 2025
+--------------------------------------------------------------------------------
+
+- New features
+    o 
+
+- Bug fixes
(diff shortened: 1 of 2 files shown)


ciju-nvidia:
RERUN

PR #2056, opened 2025-09-18
hw-mgmt: scripts: Add hw-mgmt tree dump to sysfs monitor service

felixradensky:
Adding the hw-mgmt tree dump before termination of sysfs monitor service to facilitate service debugging.

Bug: 4603973


ciju-nvidia:
RERUN

PR #2062, opened 2025-09-26
hw-mgmt: scripts: Re-instantiate mlxreg-dpu devices

ciju-nvidia:
Very rarely on some smart switches, during a random kernel boot, some of the DPUs are not initialized. mlxreg-dpu driver fails to initialize. DPU attributes will not be available.

There is a timing issue with mlx-platfom driver and mlxreg-dpu driver. By the time, mlxreg-dpu driver tries to initialize the dpu, the relevant platform i2c bus is not created. Kernel provides a mechanism called deferred probing. In most of the cases, it does help, and we will get mlxreg-dpu probing done successfully. Very rarely, deferred probing timeout exceeds and we will end up having this issue.

This is a workaround to re-instantiate mlxreg-dpu if the kernel driver initialization failed

Bug: #4228063

### DIFF
--- a/usr/usr/bin/hw-management-helpers.sh
+++ b/usr/usr/bin/hw-management-helpers.sh
@@ -989,3 +989,22 @@ get_ui_tree_archive_file()
 	esac
 	echo $ui_tree_archive
 }
+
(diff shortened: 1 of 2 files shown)


ciju-nvidia:
RERUN

PR #2070, opened 2025-09-30
hw-management: thermal: Add retry for fro ASIC temperature reading (2 commits)

sholeksandr:
- **hw-mgmt: scripts: Add logging to sync service.**
- **hw-management: thermal: Add retry for fro ASIC temperature reading**


sholeksandr:
RERUN

PR #2073, opened 2025-10-01
hw-mgmt: thermal: Fix thermal data for sn5600

sholeksandr:
Fix thermal data for sn5600.
Mask psu errors:

"psu_err" : ["direction", "present"]}

Bug: 4649699


sholeksandr:
RERUN

PR #2386, opened 2026-03-22
cherry-pick support for SN7170_LD

EliasA5:
Cherry-pick commits needed to support SN7170_LD device from https://github.com/Mellanox/hw-mgmt/tree/V.7.0050.3081


acoifmannvidia (commented):
60.1000 is frozen,

PR #2429, opened 2026-04-05
hw-mgmt: patches: Update platform driver for SPC6 platforms

felixradensky:
Ensure that driver waits for initialization of all host I2C buses before initializing i2c mux driver, to prevent the mux driver from stealing the bus number 3, reserved for host buses. This guarantees the correct mux bus numbering, as expected by user space tools.

Bug: 43251541


felixradensky:
RERUN

vadimp-nvidia (approved):

PR #2431, opened 2026-04-10
hw-mgmt: scripts: Add optimization in for I2C driver bind

sholeksandr:
In the device connection flow, replace the single 200 ms sleep and
one-time check with a polling mechanism:

- Poll every 100 ms
- Timeout after 500 ms
- Exit early as soon as the device is successfully connected

This change:

- Reduces false failures on slow driver binds
- Eliminates unnecessary delay when the attach is fast
- Improves initialization time on systems with many I2C sensors


sholeksandr:
RERUN

PR #2432, opened 2026-04-10
hw-mgmt: scripts: Add optimization in for I2C driver bind

sholeksandr:
In the device connection flow, replace the single 200 ms sleep and one-time check with a polling mechanism:

- Poll every 100 ms
- Maximum timeout increased to 400 ms
- Exit early as soon as the device is successfully connected

This change:

- Reduces false failures on slow driver binds
- Eliminates unnecessary delay when the attach is fast
- Improves initialization time on systems with many I2C sensors

### DIFF
--- a/usr/usr/bin/hw-management-helpers.sh
+++ b/usr/usr/bin/hw-management-helpers.sh
@@ -125,7 +125,6 @@ i2c_bus_max=26
 bmc_i2c_bus_max=9
 bmc_i2c_bus_offset=70
 cpu_type=
-device_connect_delay=0.2
 
 # CPU Family + CPU Model should identify exact CPU architecture
 # IVB - Ivy-Bridge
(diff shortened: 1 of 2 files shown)


sholeksandr:
RERUN

vadimp-nvidia (approved):

usr/usr/bin/hw-management-helpers.sh
```diff
@@ -551,12 +562,22 @@ connect_device()
...
+		   	local device_connect_timer=$device_connect_delay_ms
+			local step_sec
+			step_sec=$(awk "BEGIN {printf \"%.3f\", $device_connect_poll_step_ms/1000}")
 			echo "$1" "$2" > /sys/bus/i2c/devices/i2c-$bus/new_device
-			sleep ${device_connect_delay}
-			if [ ! -L /sys/bus/i2c/devices/$bus-00"$addr"/driver ] &&
-			   [ ! -L /sys/bus/i2c/devices/$bus-000"$addr"/driver ]; then
-				return 1
-			fi
+			while true
```
acoifmannvidia, line 569: does it mean we can stuck here forever?

acoifmannvidia: We have device_connect_delay_ms=400, so we should not stuck forever. -> [ "$device_connect_timer" -lt 0 ] && break

acoifmannvidia (approved):

PR #2435, opened 2026-04-13
hw-mgmt: udev: more strict match for mlxsw PCI udev rules.

sholeksandr:
Restrict the generic PCI hwmon path rules to ATTR{name}=="mlxsw" so
non-mlxsw devices under the same pattern do not trigger thermal events.
In hw-management-thermal-events.sh switch removal, run module/gearbox
symlink teardown and mlxsw-specific asic/fan cleanup only when the
hwmon name is mlxsw, aligning script behavior with the udev filter.

Bug: 4968939


acoifmannvidia:
missing RCA on the ticket.

acoifmannvidia (commented):
pls add RCA to the ticket

sholeksandr:
RERUN

acoifmannvidia (approved):

PR #2439, opened 2026-04-13
hw-mgmt: thermal: Fix fan dir logs duplicate from TC service

sholeksandr:
Failure Summary:
While testing fan direction fault scenarios, it is observed that
hw-management-tc generates both NOTICE and WARNING logs for the same FAN
direction issue, even though the NOTICE log already includes repeat
count and duration intended to prevent log flooding.

Fix: fic log id to correct

Bug: 4967076


sholeksandr:
RERUN

acoifmannvidia (approved):

PR #2443, opened 2026-04-14
hw-mgmt: thermal: Fix amb sensor logs Incorrect duration in TC service

sholeksandr:
Failure Summary:
while simulating a high temperature condition for the ambient
sensor (port_amb), hw-management-tc logs are generated indicating
threshold breach. However, the alert metadata fields "duration" and
"repeat" are not behaving as expected.

Fix: change log id on sensor max threshold validation

Bug: 4970442


sholeksandr:
RERUN

acoifmannvidia (approved):

PR #2445, opened 2026-04-14
hw-mgmt: thermal: Add "start message" marker for initial log entry

sholeksandr:
Introduce a special marker "(+)" for the first message in a log
sequence. This helps identify the point at which TC error/warning events
begin.


sholeksandr:
RERUN

PR #2446, opened 2026-04-14
hw-mgmt: scripts: Fix potential issues caused by system datetime changes

sholeksandr:
If an application relies on delay loops based on the system wall-clock
time, it may fail when the datetime is adjusted.

Fix: Replace wall-clock time.time() with CLOCK_MONOTONIC. This ensures
that NTP or manual clock updates do not disrupt or skew poll-driven
execution.


sholeksandr:
RERUN

vadimp-nvidia (approved):

PR #2447, opened 2026-04-16
hw-mgmt: scripts: Fix initial timestamps of sysfs monitor service

felixradensky:
When the sysfs-monitor service is restarted without restarting hw-management the timestamp files contain stale values from previous hw-management initialization.

The sysfs monitor should initialize its own timestamp to current uptime at start, rather than relying on potentially stale files from previous runs.

Bug: 4840120


felixradensky:
RERUN

felixradensky:
RERUN

felixradensky:
RERUN

acoifmannvidia (approved):

felixradensky:
RERUN

PR #2458, opened 2026-04-20
hw-mgmt: service: Restart sysfs monitor services when hw-mgmt service…

felixradensky:
… is restarted

Modify service unit files to make sure that sysfs-monitor and fast-sysfs-monitor services are restarted when hw-mgmt service is restarted.

Bug: 4968565


felixradensky:
RERUN

felixradensky:
RERUN

acoifmannvidia (approved):

PR #2463, opened 2026-04-27
hw-mgmgt: thermal: Fix FAN min/max speed limits for sn4700/5600

sholeksandr:
For system FANs we have attributes fanX_min/fanX_max which show FAN
max/min speed. But for systems with 2 FANs in a drawer, speeds are
differ
for front/rear fan.

The current limits do not take this into account, and the same values
are used for both rear and front fans. This causes issues when the
actual maximum fan speed and the fanX_max attribute differ by more than
threshold.

To address this, separate configurations should be used for the front
and rear fans.

This commit adds separate min/max fan limits for SN4700 and SN56x0
systems.

SN4700/SN5600
front max - 13500
front min - 4143
rear max - 12603
rear min - 3891

Bug: 4975719

### DIFF
--- a/usr/usr/bin/hw-management.sh
+++ b/usr/usr/bin/hw-management.sh
@@ -1585,8 +1585,21 @@ msn47xx_specific()
 
 	max_tachos=12
 	minimal_unsupported=1
-	echo 25000 > $config_path/fan_max_speed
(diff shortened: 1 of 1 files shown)


sholeksandr:
RERUN

vadimp-nvidia (approved):

PR #2473, opened 2026-04-28
hw-mgmt: patches: Add missing 6.1 LED driver patch

felixradensky:
The patch fixes reporting of power LED state.
This LED is controlled by CPLD and not by SW,
so special provisions are required in the LED
driver to correctly interpret the CPLD LED register value.


felixradensky:
RERUN

acoifmannvidia (approved):

felixradensky:
RERUN

PR #2482, opened 2026-04-30
hw-mgmt: bmc: Extend leakage support with ADS7924

vadimp-nvidia:
Add support for device ADS7924.

Add new json keys:
Key					When omitted
Ads7924SoftReset			Treated as true (software reset
					before programming).
Ads7924IntConfig / Slp / Acq / Pwr	0xe0 0x00 0x00 0x00 (burst at
					0x12–0x15).
Ads7924AwakeMode / Ads7924Mode		0x20 then 0x33 (awake → auto-scan). Ads7924AlarmEnable			0x0f (all four channel alarm
					enables).
Ads7924UlChN / Ads7924LlChN		ULR/LLR come from Scale and
					NormalMin/NormalMax, or
					WarningMin/WarningMax if the Normal*
					pair is not set.

What is required for threshold programming on raw I2C (Probe: false) is the same as the code enforces today: Scale, plus a low/high pair for the window — NormalMin/NormalMax or, if you omit those, WarningMin/WarningMax — plus the usual DeviceType, Bus, Address, etc.

In hw-management-bmc-a2d-leakage-config-example.json added Ads7924_configuration_note under field_reference.device_object.


vadimp-nvidia:
RERUN

vadimp-nvidia:
RERUN

vadimp-nvidia:
RERUN

sholeksandr (approved):

vadimp-nvidia:
RERUN

PR #2484, opened 2026-05-03
hw-mgmt: bmc: Add script for setting cartridge info

vadimp-nvidia:
Script should be activated only for systems equipped with cartridges.

The purpose is to copy cartridge data from the leftmost cartridge to switch board CPLD device.

Script should be triggered only on system power cycle.


vadimp-nvidia:
RERUN

vadimp-nvidia:
RERUN

felixradensky (approved):

PR #2490, opened 2026-05-05
hw-mgmt: scripts: Fix usb0 interface udev rules

felixradensky:
Previous modification of the rules resulted in second rule not beeing invoked when BMC is rebooted. Revert the name matching logic of the second rule to fix the problem. Since kernel 6.12 is not suppoted in this branch, the fix is safe.

Bug: 5011600


felixradensky:
RERUN

acoifmannvidia (approved):

felixradensky:
RERUN

PR #2500, opened 2026-05-12
hw-mgmt: thermal: drive TC version from tc_config.json

sholeksandr:
hw-mgmt no longer keys off SKU for the service version. Each platform
tc_config template carries a tc_version string;
hw-management reads it from corresponding tc_config_*.json (defaults to 2.0)
on init.


sholeksandr:
RERUN

usr/usr/bin/hw-management-start-post.sh
```diff
@@ -162,20 +176,27 @@ fi
...
-			tc_version="2.5"
-			tc_executable="hw_management_thermal_control_2_5.py"
-			;;
-		*)
-			tc_version="2.0"
-			tc_executable="hw_management_thermal_control.py"
-			;;
-	esac
+	# TC application (v2.0 vs v2.5) is selected from tc_config.json (tc_version), not SKU.
+	tc_cfg_json="$config_path/tc_config.json"
```
ciju-nvidia, line 180: Could we have a config attribute that displays the TC version? /var/run/hw-management/config/tc_version
This will be really useful to see the TC version for the platform.

felixradensky (commented):
In the subject please change s/drive/derive

usr/usr/bin/hw-management-start-post.sh
sholeksandr: User can get TC version from "systemctl status "
We don't want to add attributes which duplicating functionality. It will create additional effort for testing/developing.

sholeksandr:
RERUN

PR #2505, opened 2026-05-13
hw-mgmt: scripts: Update usb0 hotplug handling

felixradensky:
Previous udev rule modifications have broken NVOS, the rules were not invoked since in NVOS kernel 6.1.94 the interface is named usb0 from the start, so the first rule was not matched and the second rule, that relied on renaming in the first rule also failed to apply.

The usb network interface naming was broken in kernel 6.1 between versions 6.1.115 and 6.1.132 by the following commit: d9edc3428c7: net: usb: usbnet: fix name regression As a result, the interface was named ethN instead of usb0.

The problem was fixed in kernel 6.1.133 by the following commit: 07f8bdce68b: net: usb: usbnet: restore usb%d name exception for local mac addresses

This patch does the following:

 1. Adds new patch that applies the naming fix to affected kernel versions. The patch is skipped for NVOS that uses kernel version 6.1.94
 2. Updates udev rules to remove the redundant interface renaming and updates the matching criteria of the rule that runs configuration script.
 3. Removes redundant delay loop in the configuration script related to interface renaming.
 4. Updates configuration script comments and log messages

Bug: 5020389


felixradensky:
RERUN

sholeksandr (approved):

felixradensky:
RERUN

vadimp-nvidia (approved):

acoifmannvidia (approved):

felixradensky:
RERUN

PR #2526, opened 2026-05-21
hw-mgmt: thermal: Preserve FAN relax period on small PWM jumps

sholeksandr:
Small PWM changes (below FAN_RELAX_PWM_JUMP_MIN) set relax_time to 0
and overwrote rpm_relax_timestamp with current_milli_time(), cancelling
an active relax window from a prior large jump. The trend check then ran
too early against unstabilized RPM and could raise a spurious TACHO
fault.
Use max(current_milli_time() + relax_time, rpm_relax_timestamp) so small
steps do not shorten stabilization started by a larger PWM change.

Bug: 5001186


sholeksandr:
RERUN

PR #2543, opened 2026-05-27
hw-mgmt: scripts: Add support for new N6100_LD SMBIOS BOM record

felixradensky:
There's a HW bug in MP29816 voltage regulator, occasionally it fails to recognize devices at I2C addreses 0x6e-0x6f. To mitigate it, the MP29816 device at address 0x6e was moved to address 0x69.

To support new HW configuration, a new virtual VR category was created, called mp29816-6x, which is mapped to the same driver as existing mp29816, but allows creating an SMBIOS BOM record that places probematic VRs at address 0x6e.

Also N6100_LD sensors.conf was updated to support VR device at address 0x69.


felixradensky:
RERUN

felixradensky:
RERUN

vadimp-nvidia (approved):

PR #2551, opened 2026-05-28
hw-mgmt: sensors: Add scaling factor for N6100_LD power and current

felixradensky:
Add scaling factor for N6100_LD power and current definitions of LM5066I hot-swap controller.

Bug: 5035319


felixradensky:
RERUN

vadimp-nvidia (approved):

felixradensky:
RERUN

PR #2562, opened 2026-06-01
hw-mgmt: tests: Auto-install dependencies and fix offline test suites

acoifmannvidia:
## Summary
Makes `test.py --offline` self-sufficient and green again. Two related parts: (1) the test runner now installs all of its own dependencies, and (2) the offline test suites — which had drifted from the code they cover and been failing since ~2026-01-23 — are updated to match current, intentional code behavior.

**Result: all 23 offline suites (520 individual tests) pass.**

## Dependency handling (`test.py`, `requirements.txt`)
- `check_dependencies()` verifies **every** package in `requirements.txt` (not just pytest) and auto-installs the full file if anything is missing.
- `install_dependencies()` uses `stdout/stderr=PIPE` instead of `capture_output` for Python 3.6 compatibility.
- Test suites are launched via `sys.executable` so they use the same interpreter (and installed deps) as `test.py`, not a bare `python3`.
- Added `psutil` to `requirements.txt` (imported by modules under test, e.g. `hw_management_thermal_control.py`).

## Offline suite fixes (test-only; no production code changed)
- **HW_Mgmt_Logger**: repeat-finalization format is now `"<msg> (repeat=N, duration=Ns)"` (`0e8813c4`).
- **TC _exit_wait**: `ThermalManagement._exit_wait()` was refactored into shared `hw_management_lib.exit_wait()` (`8aaba0f1`).
- **TC 2.5 ASIC fread reset**: `handle_input()` now calls `_sdk_init_in_progress()`/`_read_asic_ready()` (`381c955d`); added to the harness.
- **Peripheral Updater**: chipup monitoring deduplicated to one entry per physical ASIC (`c164ecac`); `write_module_counter` early-returns for unsupported platforms (`1496432d`).
- **Module Counter**: same early-return guard; SKUs resolved as supported.
- **Redfish Client**: mock the `/run/lock` advisory lock so the test no longer needs a writable `/run/lock`.
- **Module/ASIC temperature populate**: output goes through `atomic_file_write()` (mkstemp + `os.replace`), which bypassed the `open` redirection and hit the real `/var/run`; redirected into the test's temp tree.

## Not included
The 4 ngci quality-gate checks (beautifier/spell/security/header) are unaffected and require CI-provisioned config under `build/ci/config/`, not present in a plain checkout.

### DIFF
--- a/tests/offline/hw_management_lib/HW_Mgmt_Logger/test_hw_mgmt_logger.py
+++ b/tests/offline/hw_management_lib/HW_Mgmt_Logger/test_hw_mgmt_logger.py
@@ -446,7 +446,10 @@ class TestHWMgmtLogger(unittest.TestCase):
 
         with open(self.test_log_file, 'r') as f:
             content = f.read()
-            self.assertIn("(repeat=5, duration=", content)
+            # Finalization appends a "(repeat=N, duration=Ns)" suffix to the
+            # original message (the message was seen 5 times).
+            self.assertIn("Test message (repeat=5", content)
+            self.assertIn("duration=", content)
 
     def test_23_hash_collision_handling(self):
         """Test that different messages with same ID are handled correctly"""
--- a/tests/offline/hw_mgmgt_sync/asic_populate_temperature/test_asic_temp_populate.py
+++ b/tests/offline/hw_mgmgt_sync/asic_populate_temperature/test_asic_temp_populate.py
@@ -2,7 +2,7 @@
 # -*- coding: utf-8 -*-
 ########################################################################
 # SPDX-FileCopyrightText: NVIDIA CORPORATION & AFFILIATES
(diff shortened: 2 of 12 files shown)


acoifmannvidia:
RERUN

acoifmannvidia:
RERUN

acoifmannvidia:
RERUN

acoifmannvidia:
RERUN

acoifmannvidia:
RERUN

vadimp-nvidia (approved):

acoifmannvidia:
RERUN

acoifmannvidia:
RERUN

acoifmannvidia:
RERUN

acoifmannvidia:
RERUN

PR #2570, opened 2026-06-03
hw-mgmt: bmc: netwroking: Add option to allow usb net IP address by NOS

vadimp-nvidia:
Add USB0_MANAGED_BY_NOS so BMC images can defer usb0 addressing to the NOS (e.g. SONiC sonic-usb-network-init) instead of hw-management-bmc static systemd-networkd configuration.

BMC:
- New hw-management-bmc-usb0-common.sh: read USB0_MANAGED_BY_NOS from /etc/hw-management-bmc-usb0.conf (1, yes, true).
- plat-specific-preps: when set, remove static .network unit and skip Address= rendering; sonic-usb-network-init may run.
- usb_net_config: when set, load g_ether and bring link up only; no static ip addr or networkd reload for addressing.
- HI189 hw-management-bmc-network-sonic.conf.example for SONiC BMC image recipes; document flag in network.conf and README.

Host (hw-management):
- hw-management-ifupdown.sh: on SONiC hosts (sonic_version.yml), skip ifup usb0 so the host NOS owns the interface.

Backward compatible: without USB0_MANAGED_BY_NOS, behavior is unchanged (static USB0_ADDRESS / .network as before). HI189 default still ships USB0_ADDRESS only.

Update bmc/README.md, DEVELOPER_GUIDE.md, and sonic-usb-network-init drop-in comments.


vadimp-nvidia:
RERUN

vadimp-nvidia:
RERUN

PR #2577, opened 2026-06-04
hw-mgmt: ifupdown: Restore ifquery guard and retry log wording

vadimp-nvidia:
Drop auto/hotplug ifquery -l check so usb0 stanzas without those keywords still get ifup. Revert retry log text to avoid log analyzer false positives.


felixradensky (commented):
Please restore the original code:
# Check if interface is defined in /etc/network/interfaces
if ! ifquery "$INTERFACE" >/dev/null 2>&1; then
        log_err "Interface $INTERFACE is not defined in /etc/network/interfaces"
        exit 1
fi

vadimp-nvidia:
RERUN

felixradensky (approved):

PR #2582, opened 2026-06-04
hw-mgmt: config: Add support for generic system configuration

vadimp-nvidia:
Allow new platforms to describe hw-management settings in /etc/<HID>/hw-management-platform.json instead of adding a new *_specific() function in hw-management.sh.

Split check_system() into:
- check_system_internal() for legacy board/product dispatch
- check_system_json() to load and apply the platform JSON
- check_system_bmc_redfish_login() for BMC Redfish login

Add hw-management-platform-json.sh to parse JSON using the existing BusyBox-safe hw-management-json-parser.sh. JSON can set shell variables, config files, I2C connection tables, named busses, PSU/ASIC maps, and whitelisted actions.

Add examples/hw-management-platform-example.json as schema reference for new platform bring-up.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,0 +1,36 @@
+<!-- SPDX-FileCopyrightText: NVIDIA CORPORATION & AFFILIATES -->
+
+# `examples/`
+
+Reference material for host CPU **hw-management** tooling: sample JSON,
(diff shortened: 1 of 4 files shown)


vadimp-nvidia:
RERUN

PR #2584, opened 2026-06-04
hw-mgmt: config: Add support for generic system configuration

vadimp-nvidia:
Allow new platforms to describe hw-management settings in /etc/<HID>/hw-management-platform.json instead of adding a new *_specific() function in hw-management.sh.

Split check_system() into:
- check_system_internal() for legacy board/product dispatch
- check_system_json() to load and apply the platform JSON
- check_system_bmc_redfish_login() for BMC Redfish login

Add hw-management-platform-json.sh to parse JSON using the existing BusyBox-safe hw-management-json-parser.sh. JSON can set system traits, shell variables, config files, I2C connection tables, named busses, PSU/ASIC maps, and whitelisted actions. Invalid connection table names and JSON apply errors abort boot.

Add examples/hw-management-platform-example.json as schema reference for new platform bring-up, including system section option comments and variables such as erot_count, vrot_count, mcu_count, and cartridge_count.

Add examples/README.md cataloging host-side example files.

### DIFF
--- a/examples/hw-management-platform-example.json
+++ b/examples/hw-management-platform-example.json
@@ -1,0 +1,115 @@
+{
+    "_comment": "Example platform config for /etc/hw-management-cfg/<HID>/platform.json. Remove _comment before deploy.",
+    "system": {
+        "_comment_bmc": "yes | no",
+        "bmc": "yes",
+        "_comment_cooling": "liquid | air | hybrid",
+        "cooling": "air",
+        "_comment_power_supply": "AC | DC",
+        "power_supply": "AC",
+        "_comment_power_source": "busbar | replaceable_unit",
+        "power_source": "replaceable_unit",
+        "_comment_power_supply_replaceable": "yes | no | na",
+        "power_supply_replaceable": "yes",
+        "_comment_fan_replaceable": "yes | no | na",
+        "fan_replaceable": "yes",
+        "_comment_ssd": "sata | nvme",
+        "ssd": "nvme",
+        "_comment_tpm": "yes | no",
+        "tpm": "yes",
+        "_comment_security": "PQC",
+        "security": "PQC",
+        "_comment_type": "Ethernet | IB | NvLink",
+        "type": "Ethernet"
+    },
+    "variables": {
+        "hotplug_fans": 4,
+        "max_tachos": 8,
+        "hotplug_pwrs": 0,
+        "hotplug_psus": 0,
+        "hotplug_pdbs": 0,
+        "hotplug_linecards": 0,
+        "erot_count": 0,
+        "vrot_count": 0,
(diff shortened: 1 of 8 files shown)


vadimp-nvidia:
RERUN

acoifmannvidia (approved):

vadimp-nvidia:
RERUN

vadimp-nvidia:
RERUN

examples/hw-management-platform-example.json
```diff
@@ -0,0 +1,115 @@
+{
+    "_comment": "Example platform config for /etc/<HID>/hw-management-platform.json. Remove _comment before deploy.",
```
ciju-nvidia: The config path should be /etc/hw-management-cfg/<HID>/platform.json
This will also align with devtree json.

examples/hw-management-platform-example.json
```diff
@@ -0,0 +1,115 @@
...
+        "_comment_type": "Ethernet | IB | NvLink",
+        "type": "Ethernet"
+    },
+    "variables": {
+        "hotplug_fans": 4,
+        "max_tachos": 8,
+        "hotplug_pwrs": 0,
+        "hotplug_psus": 0,
+        "hotplug_pdbs": 0,
+        "hotplug_linecards": 0,
```
ciju-nvidia, line 31: We will also require "hotplug_dpus" ?

usr/usr/bin/hw-management-platform-json.sh
```diff
@@ -0,0 +1,904 @@
+#!/bin/bash
```
ciju-nvidia, line 1: Hi Vadim,

IMHO, we should have a python based json parser script. It will be more simple and elegant. It will also reduce the number of lines of code.

vadimp-nvidia: I want to keep ability to port it to BMC and in BMC we still have limitation with json parser. 
However in latest images they add jq tool.
So, maybe we can use it.

examples/hw-management-platform-example.json
vadimp-nvidia: We have dpu_count
Not sure we need additionally hotplug_dpus

examples/hw-management-platform-example.json
vadimp-nvidia: CIju
Do you mean in
/etc/hw-management-cfg/<HID>/platform.json

vadimp-nvidia: /etc/hw-management-cfg/HI<>/platform.json

vadimp-nvidia:
RERUN

ciju-nvidia (approved):

PR #2586, opened 2026-06-05
hw-mgmt: scripts: Restrict udev event log size

ciju-nvidia:
In most of the operating systems, logrotate runs daily once via cron. If there is a faulty hardware, there can be many number of udev events and size of the udev_events.log can exceed the maximum configured size very fast. This will have an impact on the disk size. To prevent this, adding an explicit logrotate call in the udev event tracing function.  When the log size reaches 4M, an explicit logrotate call is made.

Bug #5040812

### DIFF
--- a/usr/etc/logrotate.d/udev
+++ b/usr/etc/logrotate.d/udev
@@ -1,9 +1,10 @@
 "/var/log/udev_events.log" {
-    size 1M
+    size 4M
     rotate 3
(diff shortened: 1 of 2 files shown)


ciju-nvidia:
RERUN

ciju-nvidia:
RERUN

vadimp-nvidia (approved):

PR #2600, opened 2026-06-09
hw-mgmt: bmc: leakage: Fix ADS1015 logic and json config

vadimp-nvidia:
HI189 ADS1015 leakage: use polling config (CfgRegVal 0x83) and Scale 0.002 for IIO raw counts. Skip Lo/Hi threshold I2C writes when the comparator is disabled.

Fix bind when an I2C client exists without a driver; set ti-ads1015 IIO scale to ±4.096 V after rebind so inputs above 2.048 V are not clipped to raw 2047. Align debug/read scripts with cfg_lo 0x83.


vadimp-nvidia:
RERUN

vadimp-nvidia:
RERUN

vadimp-nvidia:
RERUN

felixradensky (approved):

PR #2609, opened 2026-06-11
hw-mgmt: bmc: reset cuase: Add fixup logic for uncontrolled power cycle

vadimp-nvidia:
Problem:
BMC reset cause from SCU registers (hw-management-bmc-get-reset-cause.sh) is inconsistent after remote reboot: reset_power_on and reset_watchdog are reported interchangeably.
After aux_pwr_cycle followed by remote reboot, the CPLD reset cause (reset_aux_pwr_or_ref / reset_aux_pwr_or_fu under
/var/run/hw-management/system/) reliably indicates auxiliary power cycle, which should map to reset_power_on.
That CPLD attribute is available only after udev has populated system sysfs, which completes at the end of hw-management-bmc-boot-complete.sh - after the early get-reset-cause run in hw-management-bmc-init.

Solution:
At the end of hw-management-bmc-boot-complete.sh, when system sysfs thresholds are met, check reset_aux_pwr_or_ref or reset_aux_pwr_or_fu. If set to 1, set reset_power_on to 1 and clear all other top-level reset_* files under /var/run/hw-management/bmc/ (regular files only; domains/ subdirectory is skipped).

Corner cases:
- Plain remote reboot without aux_pwr_cycle: CPLD flag stays 0; no fixup. SCU ambiguity for that path is unchanged.
- BMC watchdog before boot-complete: fixup does not run; reset cause may still reflect watchdog.
- Boot-complete timeout: fixup does not run.
- get-reset-cause failure during init: fixup only updates existing files.

### DIFF
--- a/bmc/usr/usr/bin/hw-management-bmc-boot-complete.sh
+++ b/bmc/usr/usr/bin/hw-management-bmc-boot-complete.sh
@@ -27,6 +27,71 @@ count_entries() {
 	ls -A "$_d" 2>/dev/null | wc -l
 }
 
+# When CPLD reports auxiliary-power or firmware-update reset, prefer power_on over
+# ambiguous SCU-derived watchdog/software flags (see hw-management-bmc-get-reset-cause.sh).
+# Runs only after system sysfs is populated (boot-complete gate). HI189 uses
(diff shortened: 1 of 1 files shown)


vadimp-nvidia:
RERUN

PR #2610, opened 2026-06-11
hw-mgmt: pathes: [6.1, 6.12]: Fix hotplug bus-nr fixup on multi-segme…

vadimp-nvidia:
…nt mux

mlxplat_i2c_mux_complition_notify() indexed adapters[data->hpdev.nr - 2], which assumes hotplug devices live on the first mux segment starting at MLXPLAT_CPLD_CH1. On VMOD0009/COMEX (MSN2700) PSU/FAN hotplug buses are on the second segment (10..14), so the callback could dereference a NULL adapter and crash during i2c-mux-reg probe.

Run the fixup on every mux completion, map hpdev.nr relative to the current segment base_nr (or the first adapter's nr for regmap mux), and bounds-check the adapter index before use.

Bug: #4151613


vadimp-nvidia:
RERUN

vadimp-nvidia:
RERUN

felixradensky (commented):
While at it, can we also fix a typo in the function name, s/mlxplat_i2c_mux_complition_notify/mlxplat_i2c_mux_completion_notify ?

acoifmannvidia (approved):

PR #2613, opened 2026-06-14
hw-mgmt: services: Add software-only NVMe shutdown support for no-BMC

vadimp-nvidia:
Problem:
On systems without BMC, platform power off uses mlxplat_poweroff(), which asserts the CPLD halt bit late in kernel shutdown. SSD/aux power may drop before NVMe controllers finish controlled shutdown (SHN/flush), which can corrupt NVMe devices. There is no cpu_power_off_ready handshake on these platforms and CPLD changes are not allowed for this release.

Solution:
Add hw-management-nvme-shutdown.service and helper scripts for no-BMC systems with NVMe storage. During systemd shutdown (halt/poweroff/reboot/kexec), the service syncs block devices, requests NVMe controller shutdown via sysfs, waits for a safe state, and applies an optional software margin delay. Enable the unit from hw-management-start-post.sh only when nvme_present indicates NVMe (or NVMe controllers exist) and bmc_present/gp_bmc_presnt indicate no BMC. Skip on SATA-only and BMC-equipped platforms. ExecCondition re-checks eligibility at shutdown. Configuration is via /etc/default/hw-management-nvme-shutdown.

Testing:
Build the hw-management host package and verify scripts, default config, and unit are installed. On a no-BMC NVMe platform, confirm `systemctl is-enabled hw-management-nvme-shutdown.service` after hw-management start. On BMC or SATA platforms, confirm the unit stays disabled. Run `poweroff` or
`systemctl start hw-management-nvme-shutdown.service` on an eligible platform and check syslog for shutdown messages.

### DIFF
--- a/usr/etc/default/hw-management-nvme-shutdown
+++ b/usr/etc/default/hw-management-nvme-shutdown
@@ -1,0 +1,16 @@
+# Configuration for hw-management NVMe shutdown hook (software-only, no-BMC).
+#
+# Installed as /usr/lib/systemd/system-shutdown/hw-management-nvme-shutdown.
+# Runs in the final systemd-shutdown phase, after umount and service teardown,
+# immediately before the kernel poweroff/reboot path (pm_power_off / CPLD halt).
+#
+# Eligibility (checked at shutdown time):
+#   - NVMe block devices exist under /dev/nvme*, and
+#   - bmc_present == 0 and gp_bmc_presnt == 0 (no BMC on platform).
+#
+# NVME_SHUTDOWN_WAIT_SEC + 5 must not exceed NVME_SHUTDOWN_TIMEOUT_SEC.
+# Wait is clamped at runtime to fit the budget.
+NVME_SHUTDOWN_TIMEOUT_SEC=120
+
+# Max time to wait for NVMe controller shutdown (seconds).
+NVME_SHUTDOWN_WAIT_SEC=10
--- a/usr/usr/bin/hw-management-nvme-shutdown-condition.sh
+++ b/usr/usr/bin/hw-management-nvme-shutdown-condition.sh
@@ -1,0 +1,104 @@
+#!/bin/bash
+##################################################################################
+# SPDX-FileCopyrightText: NVIDIA CORPORATION & AFFILIATES
+# Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# Return 0 when the NVMe shutdown hook should run:
+#   - NVMe block devices are present under /dev/nvme*
+#   - platform has no BMC (software-only shutdown path)
+#
+# Return 1 to skip.
+#
+# At systemd-shutdown time /var/run/hw-management is usually gone, so a flag
+# cached at boot in /var/lib/hw-management/nvme-shutdown-enabled is used.
+##################################################################################
+
+set -euo pipefail
+
+readonly LOGGER_TAG="hw-management-nvme-shutdown"
+readonly HW_MGMT_SYSTEM=/var/run/hw-management/system
+readonly FLAG_DIR=/var/lib/hw-management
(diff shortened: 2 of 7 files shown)


usr/usr/bin/hw-management-nvme-shutdown.sh
```diff
@@ -0,0 +1,150 @@
...
+			state=$(nvme_controller_state "$ctrl")
+
+			case "$state" in
+				live|ready|resetting)
+					pending=1
+					log_msg "$name state=$state (waiting)"
+					;;
+				*)
+					log_msg "$name state=$state"
+					;;
```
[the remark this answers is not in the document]
vadimp-nvidia: pstream Linux NVMe does expose /sys/class/nvme/nvmeN/shutdown on recent kernels; the absent-sysfs path is now a safe fallback, not a guaranteed timeout on every shutdown. If your BSP strips that attribute, only sync/flush runs — worth confirming on target kernel.

vadimp-nvidia:
RERUN

vadimp-nvidia:
RERUN

usr/usr/bin/hw-management-nvme-shutdown-condition.sh
```diff
@@ -0,0 +1,105 @@
...
+		fi
+	done
+
+	return 1
+}
+
+nvme_storage_configured() {
+	local path val ctrl
+
+	path="$HW_MGMT_SYSTEM/nvme_present"
```
mshych: It doesn't exist on all systems. Should be changed to common check e.g. /dev/nvme*

usr/usr/bin/hw-management-nvme-shutdown-condition.sh
```diff
@@ -0,0 +1,105 @@
...
+		log_dbg "nvme_present=0 ($path), skip NVMe shutdown hook"
+		return 1
+	fi
+
+	for ctrl in /sys/class/nvme/nvme*; do
+		[ -d "$ctrl" ] || continue
+		return 0
+	done
+
+	log_dbg "No NVMe controllers and no nvme_present GPIO, skip hook"
```
mshych: Should be changed. No NVME GPIO on all systems

usr/usr/bin/hw-management-nvme-shutdown-hook.sh
```diff
@@ -0,0 +1,20 @@
...
+# are stopped and filesystems are unmounted, immediately before the kernel
+# poweroff/reboot path (pm_power_off / CPLD halt on no-BMC platforms).
+#
+# Argument: poweroff | halt | reboot | kexec
+##################################################################################
+
+set -euo pipefail
+
+if [ -f /etc/default/hw-management-nvme-shutdown ]; then
+	# shellcheck disable=SC1091
```
mshych, line 16: Do we need this comment?

usr/etc/default/hw-management-nvme-shutdown
```diff
@@ -0,0 +1,20 @@
...
+# systemd-shutdown overall budget; keep hook runtime below this value.
+# NVME_SHUTDOWN_WAIT_SEC + NVME_SHUTDOWN_MARGIN_SEC + 5 must not exceed this.
+# Both wait and margin are clamped at runtime to fit the budget.
+NVME_SHUTDOWN_TIMEOUT_SEC=120
+
+# Max time to wait for NVMe controller shutdown (seconds).
+NVME_SHUTDOWN_WAIT_SEC=10
+
+# Extra margin after NVMe reports safe state (seconds).
+NVME_SHUTDOWN_MARGIN_SEC=2
```
mshych: It looks that MARGIN just complicates things.
We can work just with NVME_SHUTDOWN_WAIT_SEC

mshych (commented):
Pls see comment

PR #2614, opened 2026-06-15
hw-mgmt: bmc: Fix 0009 SPC6 DTS split sund-sonic.dtsi patch hunk

vasily-vityukov-tandemg:
Correct @@ -0,0 +1,8 @@ line count so the closing }; is applied (was +1,7).


vasily-vityukov-tandemg:
RERUN

vadimp-nvidia (approved):

PR #2617, opened 2026-06-15
hw-mgmt: scripts: Update list of modules to load

ciju-nvidia:
This patch removes the module i2c-mux-regmap from
the list of modules to be loaded at the hw-mgmt
startup. This is not a required module.

Bug #5085644


ciju-nvidia:
RERUN

felixradensky (approved):

PR #2627, opened 2026-06-22
hw-mgmt: bmc: Add offline validation tests and fix test runner

acoifmannvidia:
Add offline shell-based validation tests for 10 BMC scripts under bmc/tests/, covering: json-parser, reset-cause, thermal, powerctrl, show-reset-cause, cpld-dump, boot-complete, gpio-set, generate-dump, and sysfs-validation.

Each test runs entirely offline using mktemp scratch dirs, mock binaries, and sed-patched copies of the scripts under test.

Integrate all BMC tests into tests/test.py --offline via a new run_bmc_tests() method.  Also fix several issues in the test runner:

- Run pytest suites in-process via pytest.main() instead of spawning subprocesses, avoiding an NFS rpc_wait_bit_killable hang when importing pytest from a network-mounted home directory.

- Snapshot and restore sys.modules around each pytest.main() call so test suites that replace sys.modules entries with MagicMock objects (e.g. test_hw_management_peripheral_updater) do not poison later suites (Redfish Client, TC _exit_wait).

- Fall back to a real import attempt when importlib.util.find_spec() returns None for packages on NFS-mounted site-packages, and recover gracefully when pip install returns non-zero but all packages are already importable.

- Add --skip-dep-check flag to bypass the dependency check when pip cannot reach PyPI.  Always show pip stderr on failure.

- hw-management-bmc-sysfs-validation: exit 0 (SKIPPED) when /var/run/hw-management is absent and HW_MGMT_ROOT was not explicitly overridden; the script validates live hardware and should not count as a failure on dev machines.

Result: 34/34 test suites pass, 527 individual tests green.


acoifmannvidia:
RERUN

acoifmannvidia:
RERUN

acoifmannvidia:
RERUN

acoifmannvidia:
RERUN

acoifmannvidia:
RERUN

vadimp-nvidia (approved):

PR #2629, opened 2026-06-22
hw-mgmt: kernel: making i2c-mux-reg as part of kernel

ciju-nvidia:
In some NOSs, it has been observed that there is a timing issue. The completion is only signaled from mlxplat_i2c_mux_complition_notify(), which is called by the `i2c-mux-reg` driver's probe function. That probe doesn't run within the 120-second window.

When `platform_device_register("i2c-mux-reg")` is called, if the `i2c-mux-reg.ko` module is not yet loaded, the device is registered but no probe runs. There is no driver bound to it. The kernel defers the probe until the driver module loads.

This patch marks i2c-mux-reg driver as part of the kernel. CONFIG_I2C_MUX_REG=y

Bug #5079211


ciju-nvidia:
RERUN

PR #2636, opened 2026-06-25
hw-mgmt: pathes: [6.1, 6.12]: Fix module loading/booting timing issue

vadimp-nvidia:
Problem description:
Mux probing is failed on wait_for_completion_timeout() with kernel config: CONFIG_I2C_MUX_REG=m:
mlx-platform runs early. mlxplat_i2c_mux_topology_init() registers i2c-mux-reg platform devices.
Then it is blocked until every mux completion_notify has run (including mlxplat_post_init()).
i2c-mux-reg is not loaded yet -> no driver -> mux pdevs never probe -> mux_added never reaches mlxplat_mux_num -> timeout.

With CONFIG_I2C_MUX_REG=y, the driver is already registered when pdevs are created, probes run immediately, completion fires in seconds, so no issue in this case.

So, this is a module load order / boot timing problem,

Solution:
Return -EPROBE_DEFER when mux platform data is configured but the matching i2c-mux-reg or i2c-mux-regmap platform driver is not registered yet (e.g. CONFIG_I2C_MUX_REG=m and modules-load runs after mlx-platform probe).

### DIFF
--- a/recipes-kernel/linux/linux-6.1/0116-platform-mellanox-Downstream-Defer-post-init-until-I.patch
+++ b/recipes-kernel/linux/linux-6.1/0116-platform-mellanox-Downstream-Defer-post-init-until-I.patch
@@ -9,20 +9,27 @@ callbacks complete, with spinlock, completion, 120s timeout, NULL-adapter path,
 mux teardown vs late-notifier coordination, and unconditional mux completion_notify
 on i2c-mux-reg.
 
+Pin the mux platform driver module with try_module_get() while registering mux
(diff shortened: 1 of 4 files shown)


vadimp-nvidia:
RERUN

vadimp-nvidia:
RERUN

vadimp-nvidia:
RERUN

vadimp-nvidia:
RERUN

vadimp-nvidia:
RERUN

PR #2638, opened 2026-06-26
hw-mgmt: script: Fix cpu_type init order on hw-mgmt start

sholeksandr:
On early call cpu_type reading (before hw-mgmt tree init) it
can cause log error:

/usr/bin/hw-management-helpers.sh: line 234:
/var/run/hw-management/config/cpu_type: No such file or directory

For example run iorw before hw-mgmt start

Fix: add explicit folder create on cpu_type call.

Bug: 5100761


acoifmannvidia (approved):

sholeksandr:
RERUN

PR #2644, opened 2026-06-29
hw-mgmt: thermal: add debug print for FAN speed out of range

ciju-nvidia:
Add debug print for FAN speed out of range. Print FANX_TACHO CPLD register value in case wThermalControl read abnormal FAN tacho value from hw-mgmgt fanX_speed_get

This additional log will help with debugging rare bugs and issues that are very difficult to reproduce.

Bug: 4955310


ciju-nvidia:
RERUN

acoifmannvidia (commented):
thanks but I dont want to risk the release, decline

PR #2648, opened 2026-06-29
hw-mgmt: thermal: Fix periodic report crash

ciju-nvidia:
The periodic report formatter could convert missing sensor values to "N/A" and then pass them to round(), causing type str doesn't define __round__ method.

Add safe rounding for report output so non-numeric sentinel values are preserved while numeric values are still formatted normally.

Bug #5104076


ciju-nvidia:
RERUN

acoifmannvidia:
RERUN

sholeksandr (approved):

PR #2649, opened 2026-06-29
hw-mgmt: udev rules: set TPM power state to always-on in the sysfs ph…

mshych:
…ysical device.

Set TPM power state to always-on in the sysfs physical device, in addition to the already updated power state of the tpm0 class device.

Addition to commit f7c685cd2b83.

Bug #4542714


mshych:
RERUN

vadimp-nvidia (approved):

PR #2657, opened 2026-06-30
hw-mgmt: scripts: Fix hw-management stop timeout handling

sholeksandr:
Description: hw-management services could exceed systemd stop timeout
because SIGTERM handling was delayed by long waits and thermal update loops.
Reduce the shared wait chunk, exit thermal loops as soon as the stop
event is set, and extend the TC service stop timeout to allow
graceful cleanup.

Bug 5107746, 5103562


sholeksandr:
RERUN

ciju-nvidia (approved):

PR #2658, opened 2026-07-01
hw-mgmt: thermal: Add module configuration for Tmax limitation

sholeksandr:
Add platform configs to cap module-reported Tmax so fan
control does not follow critical thresholds above the board thermal
limit.

New config in tc_config.json:

"val_max_clamp": 65000

Tmax will never exceed this value, even if module reporting bigger value.

FR: 4964651


ciju-nvidia (approved):

sholeksandr:
RERUN

PR #2662, opened 2026-07-01
hw-mgmt: scripts: Use shared helper for kernel module checks

sholeksandr:
hw-management.sh had a local is_module() implementation and
still used ad-hoc lsmod/grep checks in some paths. This duplicated
module
detection logic and could produce inconsistent matching or noisy lsmod
output.
1. Move is_module() to hw-management-helpers.sh
2. Use it for drivetemp and i2c_designware_platform checks.
3. Suppress harmless log noise like
   lsmod: ERROR: could not open '/sys/module/sunrpc/holders': No such
file or directory

Issue: 5112578


sholeksandr:
RERUN

vadimp-nvidia (approved):

PR #2664, opened 2026-07-05
Add initial support for N7100_LD platform

EliasA5:
Issue: 4797355

those commits add support for adding the mock FS according to the product name instead of the HW ID

and building on it by adding support for N7100_LD mock filesystems, where it was copied from N7170_LD and changed the vpd_data:
```diff
> git diff --no-index -- hw-management-hi194 hw-management-n7100-ld
diff --git a/hw-management-hi194/eeprom/vpd_data b/hw-management-n7100-ld/eeprom/vpd_data
index f43c935182..6e21e2fc32 100644
--- a/hw-management-hi194/eeprom/vpd_data
+++ b/hw-management-n7100-ld/eeprom/vpd_data
@@ -1,17 +1,17 @@
-Product Name:            N7170_LD
-Part Number:             920-9K24W-00L6-ES1
+Product Name:            N7100_LD
+Part Number:             920-9K51W-00L7-GS0
 Serial Number:           MT2513XZ00BK
 Base MAC Address:        6433aa117800
-Manufacture Date:        03/26/2025 13:40:15
+Manufacture Date:        07/03/2026 13:40:15
 Device Version:          16
 MAC Addresses:           64
 Manufacturer:            Nvidia
 Vendor:                  Nvidia
 SN:                      MT2513XZ00BK
-PN:                      692-9K33R-00MV-JES
+PN:                      920-9K51W-00L7-GS0
 REV:                     A1
 MFG_DATE:                5545450
-PROD_NAME:               N7170_LD
+PROD_NAME:               N7100_LD
 HW_MGT_ID:               194
 HW_MGT_REV:              16
 SW_MGT_ID:               5200
@@ -52,6 +52,6 @@ MAX_PSU:                 0
 MIN_PSU:                 0
 FACTORY_ASSMBL_PSU:      0
 PSID:                    ADAB6B3CC2284A4B84C79B340F8C77B5
-Platform Name:           x86_64-nvidia_n7170_ld-r0
+Platform Name:           x86_64-nvidia_n7100_ld-r0
 ONIE Version:            2024.08-5.3.0015-115200
 CHSUM_FIELD:             0X72C7F5B1
```


usr/usr/bin/hw-management.sh
```diff
@@ -4220,7 +4220,12 @@ case $ACTION in
 		fi
 		# TEMPORARY hw-management mockup values for HI180/HI181/HI185/HI193/HI194 in simx
 		if check_simx && [ "$sku" == "HI180" -o "$sku" == "HI181" -o "$sku" == "HI185" -o "$sku" == "HI193" -o "$sku" == "HI194" ]; then
-			tar -xzf /etc/hw-management-virtual/hwmgmt_$sku.tgz -C /var/run/
+			product=$(< /sys/devices/virtual/dmi/id/product_name)
```
ciju-nvidia: Please use the existing infrastructure.
At this point, just extract the archive and exit. No further checking

EliasA5: both N7170_LD and N7100_LD have the same sku, have to differentiate by the product name

ciju-nvidia (commented):
The archive should be named per sku, not with the product name. Please follow the existing convention for the different skus.

usr/usr/bin/hw-management.sh
ciju-nvidia: It's the other way.  There will be same product, but will have different SKUs. If there are deviations, please address it first.

PR #2675, opened 2026-07-10
hw-mgmt: patches: 6.12: refresh patch 0064 Renesas isl68137 context

vasily-vityukov-tandemg:
Refresh the linux-6.12 Renesas isl68137 patch context so it applies cleanly to the current 6.12.38 and 6.12.73 kernel source layouts. The Renesas driver logic is unchanged; only the nested patch hunk locations and index line are updated after source layout drift.


vadimp-nvidia (approved):

vasily-vityukov-tandemg:
RERUN

vasily-vityukov-tandemg:
RERUN

vasily-vityukov-tandemg:
RERUN

PR #2677, opened 2026-07-12
docs: use model numbers instead of aliases

acoifmannvidia:
Replace system aliases with model numbers / platform IDs in documentation, release notes, labels, and comments.
Changes are limited to text/comments and the approved UI label name .


sholeksandr (approved):

acoifmannvidia:
RERUN

PR #2681, opened 2026-07-13
Rebase two 6.12 patches onto 6.12.41 to unbreak SONiC kernel build

dzaguri-nv:
The SONiC Debian kernel build applies the patch series with fuzz 0. Two linux-6.12 patches failed to apply against kernel 6.12.41, breaking the sonic_bmc_main build (linux-headers source_sonic stage):

- 0064-hwmon-pmbus-isl68137: hunk 2 context was generated against a newer kernel where isl68137_probe() contains 'info->write_word_data = raa_dmpvr2_write_word_data;' before the raa_dmpvr2_2rail_nontc case. That line does not exist in 6.12.41, so the hunk only applied with fuzz 1 and the build rejected it. Regenerate the drivers/hwmon/pmbus/isl68137.c diff against 6.12.41.

- 0008-arm64-dts-aspeed-spc6-bmc: the hand-folded eMMC HS200 hunk (mmc_pwrseq node) produced an interleaved-context hunk shape that GNU patch 2.7.6 only accepts with fuzz 1 even though the context matches byte-for-byte. Regenerate the DTS diff canonically with git.

Verified: full SONiC 6.12 series (215 patches, common + aspeed/BMC tables) applies cleanly at fuzz 0 on vanilla v6.12.41, and the patched aspeed-nvidia-spc6-bmc.dts compiles to a DTB.


dzaguri-nv:
RERUN

vadimp-nvidia (approved):

PR #2683, opened 2026-07-13
hw-mgmt: scripts: Improve robustness for fanX_dir init

sholeksandr:
fan_dir and fan status are initialized from different asynchronous udev
events, so set_fan_direction() for a fan can run before its status is
ready and leave fanX_dir uninitialized.

Move set_fan_direction() and set_fan_direction_for_all_fans() into the
shared hw-management-helpers.sh library and run an additional
set_fan_direction_for_all_fans() pass at the end of
hw-management-start-post.sh, after udev has settled, to initialize any
fanX_dir that was missed. Debounce is only applied for fans whose
fanX_dir attribute is not present yet.

Bug: 5151050


ciju-nvidia (approved):

sholeksandr:
RERUN

PR #2684, opened 2026-07-13
hw-mgmt: bmc: Extend support for ADS7924 device configuration

vadimp-nvidia:
ADS7924 support:
- Program ULR/LLR window comparator from per-channel voltage thresholds (Scale + Normal/Warning min/max), 8-bit codes.
- Add optional Ads7924* register overrides (INTCONFIG, SLP/ACQ/ PWR, INTCNTRL alarm enable, AWAKE and final MODE bytes).
- Clear a stale alarm before scanning by reading INTCONFIG (0x12), per TI SBAS482.
- Default final MODE byte set to 0x80 per HW request (auto-scan with sleep + INT), still overridable via Ads7924Mode.
- Mirror the stale-alarm clear and MODE default in the ads7924-force-alarm debug tool.
- Add HI188 leakage config example (Mngm + CSM0..3, MAX1363 and ADS7924 alternatives).

Probing order:
- For Probe:true devices, bind the kernel driver first so the driver probe runs, then program registers via i2ctransfer with the driver left bound. This makes our configuration the last writer and keeps the IIO voltage sysfs links.
- Applies to ADS7924, MAX1363 and ADS1015; drops the previous unbind/program/rebind cycle that let the driver re-probe (and potentially overwrite our values) after configuration.

### DIFF
--- a/bmc/examples/hw-management-bmc-a2d-leakage-config-hi188-eample.json
+++ b/bmc/examples/hw-management-bmc-a2d-leakage-config-hi188-eample.json
@@ -1,0 +1,581 @@
+[
+  {
+    "Name": "Mngm_ADC0",
+    "_draft_note": "DRAFT from 'Leak Detection Application Notes.docx'. Fill real BMC I2C bus and ADS7924 address before deployment. Management ADC has one active sensor; channels 2-4 are marked unused with a full-scale window to suppress floating-input alarms.",
+    "NumChnl": 4,
+    "ChnlNames": [
+      "Mngm_ADC0_Ch0_Embedded_0",
+      "Mngm_ADC0_Ch1_Unused",
+      "Mngm_ADC0_Ch2_Unused",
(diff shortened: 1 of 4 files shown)


vadimp-nvidia:
RERUN

bmc/examples/hw-management-bmc-a2d-leakage-config-hi188-eample.json
```diff
@@ -0,0 +1,581 @@
+[
+  {
+    "Name": "Mngm_ADC0",
+    "_draft_note": "DRAFT from 'Leak Detection Application Notes.docx'. Fill real BMC I2C bus and ADS7924 address before deployment. Management ADC has one active sensor; channels 2-4 are marked unused with a full-scale window to suppress floating-input alarms.",
```
acoifmannvidia, line 4: what is this doc ? 'Leak Detection Application Notes.docx'

sholeksandr (approved):

PR #2698, opened 2026-07-16
hw-mgmt: scripts: Update hw-mgmt dump generation script (2 commits)

sholeksandr:
- hw-mgmt: debug: align sysfs snapshot and collect updater logs
- hw-mgmt: scripts: Update hw-mgmt dump generation script


sholeksandr:
RERUN

PR #2699, opened 2026-07-16
hw-mgmt: bmc: Add support for BMC watchdog service

vadimp-nvidia:
Motivation
On AST2700 SONiC BMC, the ABR watchdog is intended for redundant SPI boot. The BMC boots from eMMC while SPI holds the OpenBMC image. If the ABR watchdog expires, BootMCU may switch to the OpenBMC image. This is not desired on SONiC. The watchdog is armed only when the OTP fuse is programmed, so it matters on production systems.

Considered approaches
1. Disable the watchdog in U-Boot before handing off to Linux.
2. Disable the watchdog from Linux by writing the WDTA register.
3. Service the watchdog over MCTP to Bali (Caliptra) firmware.
4. Change Bali firmware to disable the watchdog instead of servicing it (avoided to limit Sonic-specific firmware divergence).

Solution
Add hw-management-bmc-wd-heartbeat as a stop-gap until SONiC picks a final approach. A systemd service starts after boot-complete and runs hw-management-bmc-wd-heartbeat.sh, which periodically sends mctp-client eid <EID> type control data 80 03 to Bali. Platform settings live in /etc/hw-management-bmc-wd-heartbeat.conf (deployed from HI189). The service no-ops when disabled or when mctp-client is missing. WD_HEARTBEAT_EID must be set after verification on target.

### DIFF
--- a/bmc/FILE_MAPPING.md
+++ b/bmc/FILE_MAPPING.md
@@ -15,8 +15,8 @@ Use this as a checklist when copying or updating files from the OpenBMC meta-nvi
 | File | Role |
 |------|------|
 | **`debian/hw-management-bmc.postinst`** | Custom **`configure`**: **`daemon-reload`**; fresh-install **`systemctl start --no-block`** (non-blocking first-boot under **`rc.local`**). See **`bmc/README.md`** § *Debian maintainer scripts*. Issue **#4992267**. |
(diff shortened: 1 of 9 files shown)


vadimp-nvidia:
RERUN

akomisar-nv (approved):

PR #2701, opened 2026-07-17
hw-mgmt: thermal: Fix minor thermal control issues.

sholeksandr:
Fix minor thermal control issues pointed by AI.

0Description:
1. Fix fan shutdown sysfs path (system/fanN_shutdown per drawer).
2. Fix DPU module sensor regex typo (dpu\d*_module).
3. Fix attention-fan steady-state PWM default assigned to wrong
variable.
4. Avoid per-poll sensor start/stop when state is unchanged.
5. Return early from PSU set_pwm when PWM decode fails.
6. Treat missing PWM read (None) separately from zero in RPM validation.
7. Log correct device name on emergency stop.
8. Guard rm_file() when file is missing.
9. Validate fan drawer name in add_fan_drwr_sensor().
10. Iterate copy of dev_obj_list in _get_dev_obj().
11. Fix _pwm_get_max() logger call.
12. Handle missing thermal/ directory during HW discovery.
13. Document legacy ASIC-ready default behavior.

### DIFF
--- a/usr/usr/bin/hw_management_thermal_control.py
+++ b/usr/usr/bin/hw_management_thermal_control.py
@@ -710,7 +710,8 @@ class hw_management_file_op:
         @param data: data to write
         """
         filename = os.path.join(self.root_folder, filename)
-        os.remove(filename)
+        if os.path.isfile(filename):
+            os.remove(filename)
 
     # ----------------------------------------------------------------------
     def get_file_mtime(self, filename):
@@ -1423,9 +1424,11 @@ class system_device(hw_management_file_op):
         @summary: main function to process device/sensor
         """
         if self.check_sensor_blocked():
-            self.stop()
+            if self.state == CONST.RUNNING:
+                self.stop()
         else:
-            self.start()
+            if self.state != CONST.RUNNING:
+                self.start()
 
         if self.state == CONST.RUNNING:
             # refreshing attributes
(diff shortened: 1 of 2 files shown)


usr/usr/bin/hw_management_thermal_control.py
```diff
@@ -1423,9 +1424,11 @@ def process(self, thermal_table, flow_dir, amb_tmp):
         @summary: main function to process device/sensor
         """
         if self.check_sensor_blocked():
-            self.stop()
+            if self.state == CONST.RUNNING:
```
acoifmannvidia, line 1427: are there other states we might miss?

usr/usr/bin/hw_management_thermal_control.py
```diff
@@ -1423,9 +1424,11 @@ def process(self, thermal_table, flow_dir, amb_tmp):
...
+                self.stop()
         else:
-            self.start()
+            if self.state != CONST.RUNNING:
+                self.start()
```
acoifmannvidia, line 1431: is there a possibility of force that we might miss?

usr/usr/bin/hw_management_thermal_control.py
sholeksandr: 3 possible states

    UNCONFIGURED = "UNCONFIGURED"
    STOPPED = "STOPPED"
    RUNNING = "RUNNING"

usr/usr/bin/hw_management_thermal_control.py
sholeksandr: 3 possible states

    UNCONFIGURED = "UNCONFIGURED"
    STOPPED = "STOPPED"
    RUNNING = "RUNNING"

acoifmannvidia (approved):

sholeksandr:
RERUN

PR #2706, opened 2026-07-20
hw-mgmt: bmc: leakage: Enable MAX1363 A2D readings on HI189

dzaguri-nv:
The MAX1363 leakage reader was a placeholder, so a board populated with a MAX1363 A2D built the runtime tree but never published any channel value.

- Implement max1363_read_channels(): read each channel's IIO input node and publish value = raw * scale, matching the other A2D readers.
- Force the MAX1363 Vdd IIO voltage_reference after the driver is rebound; otherwise the driver defaults to its internal 2.048 V reference and rails inputs above it to raw 4095, contradicting the 0.0008 V/LSB scale.
- Set the HI189 MAX1363 thresholds and monitor-burst config to the characterized platform values.


dzaguri-nv:
RERUN

vadimp-nvidia (approved):

PR #2717, opened 2026-07-22
hw-mgmt: scripts: Add support for N8100_LD virtual platform

NadavHuNV:
following ticket: [https://redmine.mellanox.com/issues/5119183]([Story #5119183](https://redmine.mellanox.com/issues/5119183): (3) [Design] - Simulation | QM6 Bringup NVOS over ChipSim.)


usr/usr/bin/hw-management-helpers.sh
```diff
@@ -330,7 +330,7 @@ check_labels_enabled()
 check_if_simx_supported_platform()
 {
 	case $vm_sku in
-		HI130|HI122|HI144|HI147|HI157|HI112|MSN2700-CS2FO|MSN2410-CB2F|MSN2100|HI160|HI158|HI166|HI171|HI172|HI173|HI174|HI176|HI179|HI180|HI181|HI185|HI193|HI194|HI199|HI200)
+		HI130|HI122|HI144|HI147|HI157|HI112|MSN2700-CS2FO|MSN2410-CB2F|MSN2100|HI160|HI158|HI166|HI171|HI172|HI173|HI174|HI176|HI179|HI180|HI181|HI185|HI193|HI194|HI199|HI200|HI201)
```
ciju-nvidia, line 333: If you are using the virtual platform emulation, it shouldn't be mentioned here.

usr/etc/hw-management-virtual/HI201/environment
```diff
@@ -0,0 +1,440 @@
+comex_voltmon1_curr1_input 5750
```
ciju-nvidia: How are you applying these values to the attributes?

usr/etc/hw-management-virtual/HI201/thermal
```diff
@@ -0,0 +1,14 @@
...
+comex_voltmon1_temp1_crit 120000
+comex_voltmon1_temp1_input 42000
+comex_voltmon1_temp1_max 105000
+comex_voltmon2_temp1_crit 115000
+comex_voltmon2_temp1_input 43000
+comex_voltmon2_temp1_max 105000
+
+cpu_pack 42625
+cpu_pack_crit 100000
+cpu_pack_max 95000
```
ciju-nvidia: The purpose of these files are for the real platform emulaion, on which certain platform emulation drivers are missing. Not for the virtual emulation. For the virtual emulation, the static archive should contain all the relevant files with values. 

See the hw-management-start-post.sh
'Create the links for the sensors which doesn't have emulation drivers'

ciju-nvidia (changes requested):

NadavHuNV:
RERUN

vadimp-nvidia (approved):

usr/usr/bin/hw-management-helpers.sh
```diff
@@ -478,6 +478,7 @@ process_simx_links()
 	# Create the attributes in thermal and environment directories of hw-management.
         dir_list="thermal environment alarm"
         for i in $dir_list; do
+                [ -r "$vm_vpd_path"/"$i" ] || continue
```
ciju-nvidia: Why is this change required here?

ciju-nvidia (commented):
Please squash the commits to a single one.

usr/usr/bin/hw-management-helpers.sh
NadavHuNV: @ciju-nvidia  because process_simx_links is called every time after we extract the tarball
and because since we put everything in the tarball and now thermal/env don't exist in /etc/hw-management-virtual/HI201
so it prints "hw-management-helpers.sh: line 486: /etc/hw-management-virtual/HI201/thermal: No such file or directory" every time
so now with this if this file doesn't exist it will just continue for the next file and exit clean

ciju-nvidia (commented):
There are few things to fix:

1) The commit headline should be 'hw-mgmt: scripts: Add support for N8100_LD virtual platform'.

2) There are several deleted files in the commit history. When you are pushing the changes to a repo, make sure that it's clean commit without any stale file changes.

3) Even though, your commit headline says, HI201, why are there changes on different archives? If the intention is to use the same PR, you can have a separate commit within the same PR

4) Have some simple explanation in the commit message. No need to use the wordings provided by AI tools. When you are adding a new platform, mention the platform specifications. Please see the other commit messages in the hw-mgmt repo.

ciju-nvidia (approved):

NadavHuNV:
RERUN

PR #2719, opened 2026-07-23
hw-mgmt: scripts: Gate fan PWM writes on ASIC readiness

sholeksandr:
Skip PWM writes while the ASIC is restarting to avoid expected,
non-functional mlxsw_minimal error messages.

Use safe ASIC configuration lookups and apply the change to both
thermal control implementations.

Issue: #5113179


sholeksandr:
RERUN

PR #2728, opened 2026-07-24
hw-mgmt: scripts: Fix warning in add_come_named_busses

ciju-nvidia:
The merge commit c424833f6 introduced a log_err()
for add_come_named_busses(). Earlier, this was a
silent return. Now there will be a message like
"unsupported cpu_type <cpu_id> for add_come_named_busses", for the cpu types which are not specifically handled. This commit is to restore the original behavior.

Bug #5170326


ciju-nvidia:
RERUN

felixradensky (approved):

PR #2730, opened 2026-07-24
hw-mgmt: patches: Add support for SPD5118 DDR5 SODIMM temperature sensor

felixradensky:
This patch adds SPD5118 driver the list of kernel 6.1 patches and enables the driver in 6.1 and 6.12 kernel configurations.


felixradensky:
RERUN

sholeksandr (approved):

PR #2733, opened 2026-07-26
hw-mgmt: scripts: Make all virtual platforms self-contained

NadavHuNV:
Fix to an issue in which platforms that were virtual used process_simx_links after .tgz extraction in order to get complementary files such as env/thermal from the virtual dir.
Keeping in start-post.sh script only the platforms that don't have archive file (.tgz).
Platforms that are virtual now have all their data in the archive (thermal,env) so that they are now self contained.


NadavHuNV:
RERUN

NadavHuNV:
RERUN

ciju-nvidia (approved):

PR #2735, opened 2026-07-26
hw-mgmt: scripts: fix sn2700 port CPLD read race on boot

NadavHuNV:
On SPC1 hw-mgmt runs `mst start` early just to read the port CPLD3 version.
nothing coordinates it against sx-kernel, so when MRSR reset lands between the mst_pci and
mst_pciconf nodes (according to dumps in redmine it's something like 30-40 ms) VSEC stays stuck for the whole boot and syncd/swss fail.

i moved the read into do_chip_up_down, which runs after sx_core is up as @vadimp-nvidia  suggested.

i tested on r-panther-23, 8 reboots. MRSR approx 20s, mst nodes 90-100s approx, no VSEC errors,
cpld version still reports correctly

https://redmine.mellanox.com/issues/5100061

### DIFF
--- a/usr/usr/bin/hw-management.sh
+++ b/usr/usr/bin/hw-management.sh
@@ -1116,21 +1116,11 @@ add_come_named_busses()
 	named_busses+=(${come_named_busses[@]})
 }
 
-start_mst_for_spc1_port_cpld()
-{
-	if [ ! -d /dev/mst ]; then
-		lsmod | grep mst_pci >/dev/null 2>&1
-		if [  $? -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


NadavHuNV:
RERUN

NadavHuNV:
RERUN

NadavHuNV:
RERUN

felixradensky (approved):

NadavHuNV:
RERUN

NadavHuNV:
RERUN

NadavHuNV:
RERUN

NadavHuNV:
RERUN

PR #2737, opened 2026-07-27
hw-mgmt: thermal: Add dummy PSU support to TC

sholeksandr:
iDummy PSU slots are present without a real I2C device
(config/psuN_is_dummy). Treat them as non-faulting in
thermal control: skip RPM reads, fan and PWM i2c writes,
suppress present faults, keep fan direction unknown, and
report them as dummy in TC status.

Bug: 4931230


sholeksandr:
RERUN

sholeksandr:
RERUN

PR #2751, opened 2026-07-31
hw-mgmt: scripts: Fix fan direction debounce races on hotplug

sholeksandr:
hw-mgmt: scripts: Fix fan direction debounce races on hotplug

Issue: Fan direction (fanX_dir) could be wrong or stale during rapid
fan hotplug / presence events.

RC: Debounce for an older event could still complete and write fanX_dir
after a newer event. Also, fan_dir bitfield bits are not cleared on
removal, so a stable reading alone could report direction for an
absent fan. Debounce success was tied to timer expiry rather than
confirmed stable samples.

Fix: Assign a generation per presence event with flock; abort or drop
writes when superseded. Re-check fanX_status before committing
direction. Require fan_debounce_counter >= 2 for a valid direction.
Shorten debounce to 750 ms and poll every 50 ms.

Bug: 5082250

### DIFF
--- a/Documentation/Chassis_Management_for_NVIDIA_Switch_Systems_with_Sysfs_rev.3.2.md
+++ b/Documentation/Chassis_Management_for_NVIDIA_Switch_Systems_with_Sysfs_rev.3.2.md
@@ -1740,7 +1740,7 @@ C2P
 
 **Node name:** `$bsp_path/config/fan_dir_eeprom`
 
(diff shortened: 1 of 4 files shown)


sholeksandr:
RERUN

vadimp-nvidia (approved):

sholeksandr:
RERUN

PR #2756, opened 2026-08-02
hw-mgmt: SPC1: Asic control - introduce degrade mode

vadimp-nvidia:
When mlxsw_minimal fails to probe on legacy SPC1 boards (VMOD0001/0002/003/0004/0005/0009), asic_chipup_completed stays 0 and SONiC pmon never starts.

After chipup retries are exhausted, count the ASIC as initialized anyway so NOS can come up. Record the state in asic_chipup_degraded for BSP bookkeeping and TC.

Do not write PWM: with no mlxsw_minimal attached, fans remain at the hardware default (100%). Thermal control exits cleanly instead of restart-looping on missing max_tachos. Other platforms keep the previous behavior.


vadimp-nvidia:
RERUN

PR #2765, opened 2026-08-04
hw-mgmt: script: Fix tc_config init order

sholeksandr:
Create tc_config.json before start-post TC selection

Description: start-post selects TC 2.0 vs 2.5 from config/tc_config.json,
but do_start() created that file after start-post. TC 2.5 platforms then
defaulted to TC 2.0. Move the copy before calling start-post.


sholeksandr:
RERUN

vadimp-nvidia (approved):

sholeksandr:
RERUN

PR #2769, opened 2026-08-05
hw-mgmt: thermal: Introduce user space PWM control

vadimp-nvidia:
On Spectrum-1 platforms (e.g. SN2700), mlxsw-minimal may fail to probe over I2C. Without I2C hwmon, PWM and tacho nodes are missing and hardware PWM stays at 100%, so thermal control cannot manage fans.

Add hw_management_pwm_control.py: a userspace daemon that reads and writes ASIC FAN registers over PCIe via mlxreg (MFCR, MFSC, MFSM, FORE) and publishes regular files under
/var/run/hw-management/thermal/ (pwm1, fan{N}_speed_get, fan{N}_speed_set, fan{N}_fault), applying config/fan_inversed the same way as thermal-events. Package as systemd unit hw-management-pwm-control.service and install it from the RPM spec.

Manual testing:
- On SN2700 with MST pciconf present, run: hw_management_pwm_control.py --once -v 10
- Confirm MFCR pwm/tacho masks and that thermal/pwm1, fan*_speed_get, and fan*_fault are updated.
- Write a duty cycle to thermal/pwm1 and verify MFSC pwm_duty_cycle follows on the next poll interval.

### DIFF
--- a/debian/hw-management.hw-management-pwm-control.service
+++ b/debian/hw-management.hw-management-pwm-control.service
@@ -1,0 +1,20 @@
+[Unit]
+Description=ASIC FAN PWM/tacho control via mlxreg (PCIe) for Nvidia systems
+After=hw-management.service
+Requires=hw-management.service
+PartOf=hw-management.service
+Documentation=man:hw-management.service(8)
+
+StartLimitIntervalSec=1200
+StartLimitBurst=15
+
+[Service]
+ExecStart=/bin/sh -c "/usr/bin/hw_management_pwm_control.py"
+ExecStop=/bin/kill $MAINPID
+TimeoutStopSec=10
+
+Restart=on-failure
+RestartSec=60s
+
+[Install]
+WantedBy=multi-user.target
--- a/debian/hw-management.hw-management-pwm-control.service
+++ b/debian/hw-management.hw-management-pwm-control.service
@@ -5,6 +5,9 @@ Requires=hw-management.service
 PartOf=hw-management.service
 Documentation=man:hw-management.service(8)
 
+# Started/stopped from hw-management-thermal-events.sh on sxcore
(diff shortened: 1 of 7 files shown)


vadimp-nvidia:
RERUN

vadimp-nvidia:
RERUN

vadimp-nvidia:
RERUN

usr/usr/bin/hw-management.sh
```diff
@@ -1247,10 +1247,18 @@ msn27xx_msb_msx_specific()
...
 	esac
-	# Connect TC data table 
+	# Connect TC data table
 	case $product in
+		MSN2700*)
+			# SN2700: mlxsw_minimal is not used; FAN PWM/tacho via
+			# hw-management-pwm-control.service (mlxreg over PCIe).
+			# Same as other minimal_unsupported platforms: chipup does
+			# not set asic_chipup_completed.
+			minimal_unsupported=1
```
sholeksandr, line 1257: If minimal_unsupported, how than you expect to read module/asic temp? If minimal not used - temperature sync should be done with hw_management_thermal_updater

sholeksandr: Also we should update asic_ready flag. If not minimal - it should be in hw_management_peripheral_updater

debian/hw-management.hw-management-pwm-control.service
sholeksandr, line 1: We already have hw-management-peripheral-updater.service which is specially designed for periodic update values from hw.

I think not good approach to add new service just to one task.

sholeksandr: In your case you added new log but not added it to generate_dump. In case of use hw-management-peripheral-updater log messages related to PWM will be routed to already existing log hw-management-peripheral-updater.log

PR #2778, opened 2026-08-06
hw-mgmt: scripts: Change the location of label dictionary

ciju-nvidia:
Python's pickle module can execute arbitrary code
during deserialization by crafting objects with
`__reduce__` methods. Any local user or compromised container/process can write a malicious pickle file to this default path. When the hw-management service (running as root) subsequently calls this script
with `--get_value`, it will deserialize the
attacker-controlled pickle file and execute arbitrary code with root privileges.

To prevent this issue, moving the dictionary cache under /var/run/hw-management/cache/, where only root user will have the write permission.

Bug #5054893

### DIFF
--- a/usr/usr/bin/hw-management-labels-maker.sh
+++ b/usr/usr/bin/hw-management-labels-maker.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 ########################################################################
 # SPDX-FileCopyrightText: NVIDIA CORPORATION & AFFILIATES
-# Copyright (c) 2023-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright (c) 2023-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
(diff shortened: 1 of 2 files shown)


usr/usr/bin/hw-management-labels-maker.sh
```diff
@@ -35,6 +35,7 @@ source hw-management-helpers.sh
 set -x
 sku=$(< $sku_file)
 ui_path=$hw_management_path/ui 
+cache_path=$hw_management_path/cache
```
sholeksandr: you can do even better - create hidden folder with more generic name ".tmp"
$hw_management_path/.tmp

ciju-nvidia: I made it as '.cache'

ciju-nvidia:
RERUN

ciju-nvidia:
RERUN

sholeksandr (approved):

PR #2780, opened 2026-08-06
hw-mgmt: scripts: Add platform emulation support for SN68XX_LD platforms

felixradensky:
This commit adds the hw-mgmt virtual static file based platform emulation support for SN6810_LD and SN6800_LD platforms.

Feature #4856469


felixradensky:
RERUN

yuliakonev:
RERUN

felixradensky:
RERUN

ciju-nvidia (changes requested):
One small change is required in hw-management-start-post.sh

Need to add HI183 and HI187 in the sku check for calling process_simx_links(). This is required because we don't provide any files with values in the static archives. This will ensure that /etc/hw-management-virtual/<HI187|HI183>/ folders will be read and populate thermal, alarm, environment directories.

```
#Create the links for the sensors which doesn't have emulation drivers
if check_simx; then
        if check_if_simx_supported_platform; then
                case $sku in
                        HI166|HI176|HI171) <=== Add HI183 | HI187 also
                                process_simx_links
```

ciju-nvidia (approved):

felixradensky:
RERUN

PR #2781, opened 2026-08-07
hw-mgmt: scripts: Change the location of label dictionary

ciju-nvidia:
Python's pickle module can execute arbitrary code
during deserialization by crafting objects with
`__reduce__` methods. Any local user or compromised container/process can write a malicious pickle file to this default path. When the hw-management service (running as root) subsequently calls this script
with `--get_value`, it will deserialize the
attacker-controlled pickle file and execute arbitrary code with root privileges.

To prevent this issue, moving the dictionary cache under /var/run/hw-management/cache/, where only root user will have the write permission.

Bug #5054893


ciju-nvidia:
RERUN

sholeksandr (approved):

PR #2784, opened 2026-08-09
hw-mgmt: sensors: Update sensor configuration files for SN5600 and SN…

felixradensky:
…5600D

This commit makes the following changes:

1. Fixes out2 current line mapping for several MP2975 ASIC power controllers
2. Keeps corresponding XDPE15284 mappings unmodified, since these alternative controllers were not deployed on real systems
3. Fixes whitespace issues, headers and comments
4. Removes redundant I2C bus definitions
5. Removes redundant ASIC temperature sensor definition

Bug: 5176781


felixradensky:
RERUN

vadimp-nvidia (approved):

PR #2791, opened 2026-08-10
hw-mgmt: patches: Align MP5926 driver with upstream mainline

vasily-vityukov-tandemg:
Sync the 6.1 and 6.12 MP5926 backports with the upstream Linux 7.0+ driver (VOUT via pmbus_read_word_data and cached VOUT_MODE exponent) and regenerate Kconfig/Makefile hunks so they apply cleanly on 6.1.123 and 6.12.41.


vasily-vityukov-tandemg:
RERUN

vasily-vityukov-tandemg:
RERUN

PR #2795, opened 2026-08-11
hw-mgmt: bmc: Disable ABR watchdog heartbeat service by default

vasily-vityukov-tandemg:
Motivation
hw-management-bmc-wd-heartbeat was added as a stop-gap to service the AST2700 ABR watchdog over MCTP (f8ba502). The watchdog is armed only when the OTP fuse is programmed, so enabling the service by default is unnecessary on unfused / lab systems and can start mctpd / mctpirot0 work that is not required.

Solution
Package hw-management-bmc-wd-heartbeat.service disabled by default, matching the host hw-management-sync pattern:
  - debian/rules: drop it from the dh_systemd_enable allow-list; add a separate dh_systemd_enable --no-enable for this unit.
  - debian/hw-management-bmc.postinst: omit it from the fresh-install systemctl start --no-block list. The unit, script, and conf remain installed. Operators enable when ABR must be serviced:
  systemctl enable --now hw-management-bmc-wd-heartbeat.service

BUG: #5200030

(cherry picked from commit 83e5d7e00a7028e4f03a6f32d3bc69d789e6de66)

### DIFF
--- a/debian/hw-management-bmc.postinst
+++ b/debian/hw-management-bmc.postinst
@@ -91,14 +91,9 @@ case "$1" in
                 # leaves HW-MGMT unstarted.  Suppress systemctl chatter on
                 # success (>/dev/null 2>&1); real failures hit the WARNING below.
                 #
-                #   wd-heartbeat        stop-gap ABR-watchdog heartbeat to Bali
-                #                       (Caliptra) over MCTP.  Independent of the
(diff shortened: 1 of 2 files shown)


vadimp-nvidia (approved):

vasily-vityukov-tandemg:
RERUN

PR #2799, opened 2026-08-11
hw-mgmt: scripts:  Preserve TC run state across hw-management restart

sholeksandr:
Description: After restarting hw-management, thermal control stayed on
the previous version because PartOf= stops TC before start-post can see it
was active, and unit-path parsing from systemctl status failed once systemd
warned the unit had changed on disk. Save "started" to

/var/run/.hw-management-tc-state on TC stop (and on direct hw-management.sh
stop), restore it in start-post, and resolve the unit via FragmentPath.

Bug: 5200160

### DIFF
--- a/debian/hw-management.hw-management-tc.service
+++ b/debian/hw-management.hw-management-tc.service
@@ -16,5 +16,10 @@ TimeoutStopSec=10
 Restart=on-failure
 RestartSec=60s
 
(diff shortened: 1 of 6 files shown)


sholeksandr:
RERUN

vadimp-nvidia (approved):

PR #2803, opened 2026-08-12
hw-mgmt: scripts: Avoid empty log rollover on logger startup

sholeksandr:
Logger startup was rotating the log whenever the default size limit
differed from the service cap, which created empty rotated files.
Apply the new limits without rotating unless the active log already
exceeds the configured size.

Bug: 6557317, 5191631


sholeksandr:
RERUN

sholeksandr:
RERUN

felixradensky (commented):
Bug number in commit message is not correct

felixradensky (approved):

felixradensky (approved):

PR #2818, opened 2026-08-17
hw-mgmt: bmc: align A2D leakage config JSON format

dzaguri-nv:
Unify the HI189 leakage config notation with the OpenBMC platform configs: same key order, comment style and channel naming.

Remove the per-channel CfgReg/CfgRegVal entries from the ADS7924 Channels[] array. The configuration script never reads them; the window comparator codes are derived from the Types[] NormalMin/NormalMax and Scale, so the values were inert.

Thresholds, scales, buses, addresses and the MAX1363 and ADS1015 device blocks are unchanged.


dzaguri-nv:
RERUN

PR #2824, opened 2026-08-17
hw-mgmt: scripts: Fix updater service stop timeout

sholeksandr:
Issue: hw-management-thermal-updater.service failed with result 'timeout' (SIGKILL after stop-sigterm).

RC: CPython PEP 475 restarted blocking sx_core sysfs reads, so SIGTERM did not take effect within TimeoutStopSec=5. MAINPID was also a sh -c wrapper, and InterruptedError was treated as a normal read error.

Fix: Raise ShutdownRequested from the signal handler, do not swallow InterruptedError, ExecStart the Python process as MAINPID, and raise TimeoutStopSec to 15s. atomic_file_write cleans tmp files on shutdown without wrapping it as Exception.

bug: 5221586

### DIFF
--- a/debian/hw-management.hw-management-peripheral-updater.service
+++ b/debian/hw-management.hw-management-peripheral-updater.service
@@ -8,9 +8,9 @@ StartLimitIntervalSec=1200
 StartLimitBurst=5
 
 [Service]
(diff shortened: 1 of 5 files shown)


sholeksandr:
RERUN

vadimp-nvidia (approved):

PR #2825, opened 2026-08-17
hw-mgmt: patches: Fix SN6600_LD SODIMM temperature sensor driver

felixradensky:
SN6600_LD was erroneously using JC42 SODIMM temperature sensor driver, which is not suitable for DDR5 memories. This patch switches the platform to use SPD5118 driver instead.

Bug: 5215587


felixradensky:
RERUN

acoifmannvidia (approved):

PR #2828, opened 2026-08-18
hw-mgmt: scripts: Generate kernel module dependencies only if necessary

felixradensky:
Generating kernel module dependencies each time hw-mgmt starts is unnecessary and time consuming. Regenerate them only if module dependencies database does not exist or is older than some of the kernel module files.

Bug: 5204171


felixradensky:
RERUN

vadimp-nvidia (approved):

PR #2830, opened 2026-08-18
hw-mgmt: scripts: Generate kernel module dependencies only if necessary

felixradensky:
Generating kernel module dependencies each time hw-mgmt starts is unnecessary and time consuming. Regenerate them only if module dependencies database or module alias database do not exist or older than some of the kernel module files.

Bug: 5204171


vadimp-nvidia (approved):

felixradensky:
RERUN

PR #2837, opened 2026-08-19
hw-mgmt: patches: 6.1-6.12: Update kernel patches for mp2855

vadimp-nvidia:
The MP2855 dynamically changes its VOUT over-voltage and under-voltage fault thresholds depending on the operating point.

Currently the driver reads the OV/UV threshold selectors only once during probe and stores the corresponding offsets in ov_off[] and uv_off[]. In addition, the PMBus core normally caches limit values.

As a result, inX_crit and inX_lcrit can become stale when the MP2855 changes its VID/load-dependent threshold.

Mark the VOUT OV/UV fault limits as dynamic using pmbus_set_update() and read the current threshold selector whenever the limit is requested.

Remove the probe-time threshold cache since the offsets are now obtained dynamically.

Also fix the missing assignment of the return value from mp2855_rails_vout_thresholds_get() in the original probe code.

FR https://redmine.mellanox.com/issues/5221265
Follow up https://redmine.mellanox.com/issues/5203078


vadimp-nvidia:
RERUN

felixradensky (approved):

PR #2839, opened 2026-08-19
hw-mgmt: scripts: Add bash completion for hw-management.sh

sholeksandr:
Operators typing hw-management.sh on the switch have to remember
action names (start, stop, chipup, chipdown, ...). Add a bash
completion drop-in under /etc/bash_completion.d/ so Tab completes
those actions and, for chipup/chipdown, ASIC index and PCI bus id.

Install the file from Debian packaging. Suggest bash-completion
so the drop-in is sourced.


sholeksandr:
RERUN

acoifmannvidia (approved):

PR #2847, opened 2026-08-20
hw-mgmt: scripts: Add LED software/hardware control owner indication

sholeksandr:
Userspace needs to know whether a chassis LED is controlled by software,
hardware, or both.

Expose the LED control owner through:
./led/led_<name>_control

The LED owner can optionally be defined in config/led_control_type.

The default owner is led_hw_sw.

Possible options:

led_sw - LED state is controlled by the OS from userspace.
led_hw - LED state is automatically controlled by the hardware (CPLD)
state machine.
led_hw_sw - LED state is automatically controlled by the hardware, but
can also be controlled by userspace.


sholeksandr:
RERUN

PR #2855, opened 2026-08-21
hw-mgmt: scripts: Align VR DPC flash with OpenBMC shared logic

vasily-vityukov-tandemg:
Drop the unused i2c_cmd expected-exit 0 argument and run i2cset via argv.
Accept HI###/hi### HID. Compare DPC_REVISION_ID_PAGE and PAGE2_MAX_REG
with cfg_to_int so decimal config values are not parsed as hex.
Read the CSV revision field via process substitution so a failed gate actually aborts.
Restore Infineon write-protect on early exit; skip or confirm before clearing WP.
Abort batch update if JSON validation fails.

### DIFF
--- a/usr/usr/bin/hw-management-read-vr-model-version.sh
+++ b/usr/usr/bin/hw-management-read-vr-model-version.sh
@@ -104,24 +104,18 @@ log_message()
     echo "[$level] $message"
 }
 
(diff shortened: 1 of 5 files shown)


vasily-vityukov-tandemg:
RERUN

vasily-vityukov-tandemg:
RERUN

vadimp-nvidia (approved):

PR #2860, opened 2026-08-23
hw-mgmt: tc: Replace service-file patching with a launcher script

acoifmannvidia:
Reverts 8f04e45f (Preserve TC run state across hw-management restart) and fixes the root cause properly.

Problem: hw-management-tc.service shipped a hardcoded v2.0 ExecStart. hw-management-start-post.sh patched the unit file on every boot and scheduled a daemon-reload + restart to pick up the new binary.  The state-file mechanism in 8f04e45f worked around the resulting races but left the design ownership wrong: hw-management was deciding which TC binary to run instead of the TC service itself.

Fix: introduce hw_management_tc_launcher.sh as ExecStart.  The launcher reads tc_version from tc_config.json (written by set_config_data before hw-management.service becomes active) and exec()s the matching Python binary directly, making Python the systemd MAINPID.  Because the right binary is chosen on every start, no unit-file patching, daemon-reload, or restart is needed.

Changes:
- hw-management-tc.service: ExecStart -> launcher; drop ExecStop kill wrapper and state-file ExecStartPost/ExecStop hooks
- hw_management_tc_launcher.sh: new launcher (exec, not sh -c)
- hw-management-start-post.sh: remove get_tc_version(), service-file sed patching, FragmentPath lookup, consume_tc_saved_state() call, tc_should_reload/restart/is_active variables and daemon-reload step
- hw-management-helpers.sh: remove tc_state_file and consume_tc_saved_state()
- hw-management-tc.service.8: document launcher approach
- test_hw_management_start_post_tc.py: replace state-file and reload tests with launcher version-selection tests (17 tests, all pass)

Bug: 5200160

### DIFF
--- a/debian/hw-management.hw-management-tc.service
+++ b/debian/hw-management.hw-management-tc.service
@@ -1,5 +1,5 @@
 [Unit]
-Description=Thermal control service (ver 2.0) of Nvidia systems
+Description=Thermal control service of Nvidia systems
 After=hw-management.service
 Requires=hw-management.service
 PartOf=hw-management.service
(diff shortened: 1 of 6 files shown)


felixradensky:
@greptileai review

PR #2864, opened 2026-08-25
hw-mgmt: thermal: Fix PWM stuck high after tacho error recovery

sholeksandr:
Issue: After a simulated fan tacho error, PWM correctly rose to the
fan_err.tacho dmin (65%). After restoring the real tacho sysfs, PWM
stayed at 65% and never recovered. Seen with TC 2.5.

RC: TC 2.5 stores PWM as float percent after a 0..255 sysfs round-trip,
so read PWM can differ from set PWM by <1% (65.0 -> 166/255 -> 65.1).
_validate_rpm required exact pwm_curr == pwm_set to treat speed as
settled. After the tacho-fault PWM bump that equality never became
true, so the cached tacho fault stuck and PWM stayed high.

Fix: Treat PWM as settled when |read-set| < 1. Add a unit test that a
50 vs 50.2 mismatch can still clear a cached tacho fault.

Bug: 5232793

### DIFF
--- a/tests/offline/test_hw_management_fan_validate_rpm.py
+++ b/tests/offline/test_hw_management_fan_validate_rpm.py
@@ -151,6 +151,24 @@ def test_validate_rpm_read_pwm_mismatch_uses_cached_false(module_name, tc_mod, t
 
 
 @pytest.mark.parametrize("tacho_idx", [1, 2])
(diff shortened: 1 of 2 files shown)


sholeksandr:
RERUN

sholeksandr:
RERUN

vadimp-nvidia (approved):

acoifmannvidia (approved):

PR #2867, opened 2026-08-25
hw-mgmt: thermal: SPC1: Add fallback in case if I2C ASIC driver failure

vadimp-nvidia:
This change is only for SPC1 based systems. After mlxsw_minimal chipup fails, thermal sysfs PWM is missing and TC cannot force full speed. Set PWM to 100% via mlxreg MFSC.

The reason of these failures might be HW issue: system before re-spin, bad serial cable. Or in some rare case problem with I2C communication from CPU to SPC1.
In case the problem with the serial cable - the suggestion to replace the cable by cables recommended by Nvidia.

It was agreed on the following SW approach:
In case issue happened after cold reboot - just reboot system again.
In case issue happened after warm reboot - need to change code and set PWM to 100%, so it will stay at full speed with no changes.

Bug #4966961

### DIFF
--- a/tests/offline/test_hw_management_chipup_pwm_fail.py
+++ b/tests/offline/test_hw_management_chipup_pwm_fail.py
@@ -1,0 +1,260 @@
+#!/usr/bin/env python3
+################################################################################
+# SPDX-FileCopyrightText: NVIDIA CORPORATION & AFFILIATES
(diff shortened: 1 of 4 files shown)


sholeksandr (approved):

vadimp-nvidia:
RERUN

vadimp-nvidia:
RERUN

PR #2870, opened 2026-08-26
hw-mgmt: scripts: Create DPU paths before voltmon links

ciju-nvidia:
DPU voltmon udev events can arrive before the DPU_READY event creates dpuX directories. Link creation then silently fails, causing missing voltmon2 attributes. Create the DPU environment, alarm, and thermal directories before processing HI160 DPU voltmon links.

Bug #5183838


ciju-nvidia:
RERUN

vadimp-nvidia (approved):

PR #2874, opened 2026-08-27
hw-mgmt: scripts: Fix ASIC PCIE bus mapping for SN6800_LD platform

felixradensky:
Add ASIC PCIE bus mapping for HI187 HW ID. The mapping for HI188 will be added later, when actual HW is available.

Bug: 5238405


felixradensky:
RERUN

sholeksandr (approved):

PR #2876, opened 2026-08-28
hw-mgmt: scripts: Collect sx_sdk version also in compact dump

sholeksandr:
Compact dumps had no SDK version record. Move the sx_sdk version
collection out of the non-compact branch and lower its timeout to 5s.


sholeksandr:
RERUN

PR #2878, opened 2026-08-31
hw-mgmt: scripts: Fixing gpio offset for Fire-Range

ciju-nvidia:
Fire-Range support was merged as part of SN6800_LD platform commit. Correcting the config flash reset gpio reset offset

FR #4940767


ciju-nvidia:
RERUN

vadimp-nvidia (approved):

PR #2885, opened 2026-09-02
hw-mgmt: scripts: Add support for SN6600 (initial commit)

sholeksandr:
Add support for HI186 Nvidia switch.
HI186 is a 102.4Tbps switch based on Nvidia SPC-6 ASIC equipped with 64
OSFP ports supporting 100Gbps - 800Gbps speeds, 1+2 service ports

Both equipped with:
- Air-cooled with 5 fan units.
- 2 + 2 redundant 2700W PSUs.
- System management board based on AMD CPU with secure-boot support.


sholeksandr:
RERUN

PR #2892, opened 2026-09-04
hw-mgmt: scripts: add function-call traces for start and udev debug

sholeksandr:
Sparse print_function_call traces in hw-management start/stop,
chipup, chassis/thermal udev handlers, and helpers. Log skip/fail and
lifecycle points to /var/log/hw-mgmt.trace.log without per-attribute spam.
Also reset connect_table and named_busses in check_system so restart does
not connect devices twice.

Board: all


sholeksandr:
RERUN

acoifmannvidia (approved):

PR #2898, opened 2026-09-07
hw-mgmt: scripts: Fix SN2700 port CPLD read race on boot

NadavHuNV:
Cherry-pick of f9d72a3 from V.7.0070.1000_BR (shipped in V.7.0070.1014).
Applies cleanly, no conflicts.

On SPC1 hw-mgmt runs mst start early just to read the port CPLD3 version, and
nothing coordinates it against sx-kernel. When the MRSR reset lands between the
mst_pci and mst_pciconf nodes (31-64 ms apart), VSEC stays stuck for the whole
boot and syncd/swss fail.

Moved the read into do_chip_up_down, which runs after sx_core is up. The restart
path gets the same read, since chip-up returns early there.

Bug: 5100061

### DIFF
--- a/usr/usr/bin/hw-management.sh
+++ b/usr/usr/bin/hw-management.sh
@@ -1139,21 +1139,11 @@ add_come_named_busses()
 	named_busses+=(${come_named_busses[@]})
 }
 
-start_mst_for_spc1_port_cpld()
-{
-	if [ ! -d /dev/mst ]; then
-		lsmod | grep mst_pci >/dev/null 2>&1
-		if [  $? -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


felixradensky (dismissed):
Please follow the convention to mention bug numbers in commit message, replace bug URL by:
Bug: 5100061

Also, please remove this sentence from commit message: Original validation: 8 reboots on r-panther-23, no VSEC errors, CPLD version
still reported."

felixradensky (approved):

NadavHuNV:
RERUN

NadavHuNV:
RERUN

PR #2899, opened 2026-09-07
hw-mgmt: scripts: Prefer HTTP for NOS-BMC Redfish communication

sholeksandr:
Use HTTP (port 80) by default for internal CPU-BMC Redfish. Retry
HTTPS only if HTTP is unreachable, so mixed BMC firmware still works
while HTTPS is being removed. Internal HTTPS did not verify
certificates and would break if mTLS is enabled on external interfaces.

FR: 5064006


sholeksandr:
RERUN

PR #2900, opened 2026-09-07
hw-mgmt: sensors: ignore PSU hotspot temperatures on SN5640 and SN5610.

felixradensky:
PSU FW on SN5640 (HI171) and SN5610 (HI172) reports the ambient thresholds (OTW 63C / OTP 68C) for all three PSU temperature sensors, while the primary and secondary hotspot sensors are rated 105C/120C and 120C/125C respectively. The dps460 driver has no PMBus page support, so even FW that exposes per-page thresholds cannot be read correctly, and the hotspot sensors raise false alarms once they pass 68C.

Ignore temp2/temp3 for the PSU chips in sn5640_sensors.conf, shared by both SKUs, and skip the psu*_temp2 attribute group in the hw-management sysfs tree, where the same wrong threshold shows up as psu*_temp2_max and as an asserted psu*_temp2_max_alarm. The alarm bit is computed by the PSU itself, so it cannot be corrected on the host side. The teardown path needs no change, as check_n_unlink tolerates absent links.

Bug: 5125814

### DIFF
--- a/usr/etc/hw-management-sensors/sn5640_sensors.conf
+++ b/usr/etc/hw-management-sensors/sn5640_sensors.conf
@@ -379,6 +379,10 @@ chip "xdpe1a2g7b-i2c-*-6e"
     ignore curr6
 
 # Power supplies
+# PSU FW reports the ambient thresholds (OTW 63C / OTP 68C) on all three temperature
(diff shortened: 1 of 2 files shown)


felixradensky:
RERUN

acoifmannvidia (approved):

PR #2902, opened 2026-09-07
hw-mgmt: kernel: patch 6.1/6.12: Fix duplicate /dev/i2c-N node creation

felixradensky:
i2c_dev_init() registers a bus notifier and then walks the adapters that are already registered. An adapter registered in that window is attached by both paths, and the second cdev_device_add() fails with -EEXIST:

  sysfs: cannot create duplicate filename
  '/devices/platform/mlxplat/i2c_mlxcpld.1/i2c-1/i2c-2/i2c-dev/i2c-2'
  kobject: kobject_add_internal failed for i2c-2 with -EEXIST

This is seen on systems where the mlxplat mux segments are created asynchronously while i2c-dev is being loaded by systemd-modules-load.

Make the attach path idempotent by rejecting an adapter that already has an i2c_dev while holding i2c_dev_list_lock.

Same fix for both kernels, i2c-dev.c is identical in 6.1 and 6.12 in this area.

Patch is pending upstream review on linux-i2c.

Bug: 5037522


felixradensky:
RERUN

felixradensky:
RERUN

ciju-nvidia (approved):

acoifmannvidia (approved):

PR #2908, opened 2026-09-08
hw-mgmt: thermal: Make FAN relax time configurable per platform

sholeksandr:
TC compare  expected and current FAN speed, but fans cannot stabilize
immediately (HW specific). Dut to this HW specific, TC waits for
the FAN speed to stabilize. The default stabilization time is 10s, but
for some FAN types it is not enough. If the FAN speed is not stabilized
at timeout, TC can raise an error condition.

On SN2201(M) FANsi stabilisation is slower than usual, so the default
timeout (10sec) is not enough.

This commit adds a configurable "relax_time" parameter and sets it to
20s for msn2201 / msn2201m.

Bug SW: 5245926


sholeksandr:
RERUN

sholeksandr:
RERUN

acoifmannvidia (approved):

PR #2915, opened 2026-09-09
hw-mgmt: thermal: Limit chipup PWM fallback to SPC1 warm reboot

acoifmannvidia:
The MFSC full-speed fallback is only for Spectrum-1 after a COMEX/CPU warm reboot. Skip it on other platforms and after a cold power-cycle, where recovery is another reboot.

After mlxsw_minimal probe failure thermal/pwm1 is not created, so program MFSC via mlxreg. Do not skip on pwm_control and do not use the sysfs PWM path.

Bug #4966961

(cherry picked from commit 82cbe8290590290e0b34f8a9f4468c87f3b4d249)

### DIFF
--- a/usr/usr/bin/hw-management-helpers.sh
+++ b/usr/usr/bin/hw-management-helpers.sh
@@ -1919,37 +1919,101 @@ get_asic_mlxreg_dev()
 	return 1
 }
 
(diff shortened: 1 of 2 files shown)


sholeksandr (approved):

acoifmannvidia:
RERUN

PR #2917, opened 2026-09-11
hw-mgmt: sensors: Fix MP2975 DDR voltage regulator current labels

felixradensky:
The DDR voltage regulator of SN6600_LD/SN68XX_LD is an mp2975 device driving two rails. Its curr2 attribute is the output current of the first rail, but was named after the second one, while curr5, the output current of the second rail, was ignored. The remaining current attributes are the per phase currents of the two rails. Label the output current of each rail after the rail it belongs to, as already done for the voltages and the powers.

Bug: 5269629


felixradensky:
RERUN

felixradensky:
RERUN

ciju-nvidia (approved):

acoifmannvidia (approved):

PR #2920, opened 2026-09-11
hw-mgmt: scripts: Fix reset attributes counter initialization

felixradensky:
The regio event handler initializes reset_attrr_count, while the counter incremented by check_reset_attrs() is reset_attr_count, so the counter is never initialized and the misspelled variable is never read.

It happens to work today because every udev event is handled by a separate process, in which an unset variable evaluates to zero. Should the handler ever run twice in the same process, the counter would keep climbing past reset_attr_num and reset_attr_ready would never be set.

Read reset_attr_num only if the configuration attribute exists. It is set by all the currently supported systems, but a system leaving it out made the handler complain about the missing file and then about a missing operand in the comparison, once per reset attribute.

Bug: 5269665

### DIFF
--- a/usr/usr/bin/hw-management-chassis-events.sh
+++ b/usr/usr/bin/hw-management-chassis-events.sh
@@ -1171,8 +1171,12 @@ if [ "$1" == "add" ]; then
 		$led_path/led_"$name"_state
 	fi
 	if [ "$2" == "regio" ]; then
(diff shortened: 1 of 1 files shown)


felixradensky:
RERUN

acoifmannvidia (approved):

sholeksandr (approved):

felixradensky:
RERUN

PR #2924, opened 2026-09-15
hw-mgmt: scripts: Run ifup from a service and retry on lock conflicts

felixradensky:
usb0 was brought up from a udev RUN rule that retried ifup five times with 2 second delays. ifupdown2 serializes every run on a non-blocking flock() and bails out immediately when the lock is busy, so those retries only help when the other user releases it within that window. The networking service takes the lock three times during start - ifup --allow=mgmt, ifup -a and its hotplug phase - and holds it for as long as its ifup -a takes, which can exceed the 8 seconds this loop could afford while blocking a udev worker. Retrying on both sides then never converges. The failure was also invisible, since udev discards the output of RUN programs, so the reason ifup gave up was lost.

Move the ifup out of the udev worker into hw-management-ifupdown@.service, pulled in by the interface device unit through SYSTEMD_WANTS and ordered after networking.service. That ordering removes the boot race, as systemd
delays the unit until all three ifup phases have run. It does not make the conflict impossible: ordering only applies to units started in the same transaction, so nothing protects the interface from appearing while the networking service is reloaded, which runs ifreload -a and takes the same lock. With no udev worker to block, the retries can now outlast such a run:
a lock conflict, recognized from the ifup output, is retried until IFUP_TIMEOUT, while failures unrelated to the lock keep the previous short retries. Each run is in turn bounded by what is left of that budget, or by IFUP_MIN_TIMEOUT when little of it is left, so that a stalled hook or DHCP client cannot keep the script running until systemd stops the service with nothing reported.

Capture the ifup output (-v, plus -d where the installed tool supports it) and report on failure the error lines, the lock owner with its run time and command line, the other ifupdown instances, the link state and ifquery -c. The last one tells a lost race apart from a real configuration error, as the networking service brings up allow-hotplug interfaces itself and may have configured usb0 already.

Should an old rules file still invoke the script from udev after an upgrade, the ACTION variable udev exports restores the original short budget, so a udev worker is never blocked for the full timeout.

Bug: 5225696


sholeksandr (approved):

felixradensky:
RERUN

felixradensky:
RERUN

laviasaf:
RERUN

acoifmannvidia (approved):

PR #2926, opened 2026-09-15
fix: improve kernel patch descriptions

dzaguri-nv:
Fixes for:

Descriptions removed from two patches that already had them (regression)
0061-platform-mellanox-mlxreg-io-Increase-max-supported-a.patch — the four-line rationale ("Newer NVIDIA/Mellanox platforms expose more mlxreg-io register attributes than the current limit of 128 allows… Raise MLXREG_IO_ATT_NUM from 128 to 250") is replaced by the one-liner "Increase the number of supported mlxreg-io attributes."
0062-Add-support-for-NXP-s-PCF85053A-RTC-chip.patch — the description is removed entirely; the patch now has no body at all.
A functional change shipped with no description update
0046-platform-mellanox-nvsw-bmc-Add-system-control-and-mo.patch changes 14 mlxreg_core_data register descriptors in 7.0070.1016 — 13 in nvsw-bmc-hid162.c, 1 in nvsw-host-l1.c. The .bit field is corrected from the mask's highest bit to its lowest bit + 1 (affecting bios_status, bios_status_ro, cpu_spi_ctrl, psu_qty, asic_clk_brd_hw_id, config1). This is a real bug fix — for example psu_qty previously read back 2147483648 instead of 2 on a two-PSU system.

Missing author metadata
0070-nvme-pci-try-flr-on-controller-disable-failure.patch has only a Subject: line — no From:, no Date:, no Signed-off-by:, no Upstream-Status:. The description body itself is good and detailed; it just has no attributable author. Please have it regenerated with git format-patch so the headers are tool-generated.

Internal-only reference
0068-iio-adc-max1363-Add-ability-to-set-scale-of-the-ADC.patch cites "Fixes JIRA" as its justification. Community reviewers cannot open that link. Please replace it with a self-contained sentence describing the hardware problem. Minor formatting note on the same patch: the "Upstream-Status: Pending" line sits above the mbox "From " header instead of inside the commit message.


sholeksandr (approved):

dzaguri-nv:
RERUN

dzaguri-nv:
RERUN

dzaguri-nv:
RERUN

dzaguri-nv:
RERUN

PR #2943, opened 2026-09-17
kernel patches integration with sonic bmc minor upgrade to 6.12.94

dzaguri-nv:
Rebase SONiC BMC kernel patches 6.12.41 -> 6.12.94.

0001 mctp: dropped hunks now in base - mctp_frag_queue() skb ownership, mctp_getsockopt() -ENOPROTOOPT, SOCK_RCU_FREE in mctp_sk_hash() (we were setting it twice), mctp-i2c rx skb leak, mctp-i2c tx flow-state ordering. All NVIDIA deltas kept; route->dst refactor is rename-only. Real change, since mctp_frag_queue() now frees the skb on error: SOM caching moved before the call (skb_unshare() can free it even on success), and mctp_report_rx_sequence_error() takes net_device + a stack copy of the header instead of the freed fragment. Its "fall back to the failed fragment when reasm_head is NULL" path is gone - unreachable, the call site is the reasm_head != NULL branch.

0004 vdm-pcie: #ifdef -> #if IS_ENABLED() already in base, dropped. Left: <linux/kconfig.h> + MODULE_LICENSE/DESCRIPTION.

0006 jtag: context only, base adds <linux/iopoll.h>.

0010 ftgmac100: comment-only now. Base resets every Aspeed SoC, not just RMII, and adds phy_reset(priv->sgmii) for AST2700 - covers the SGMII RX stall. Comment kept so it is not narrowed back per PHY mode.

kconfig: add CONFIG_ASPEED_ESPI=y. CONFIG_AST2700_ESPI is not a real symbol, so the eSPI slave driver was never built. Only runtime behaviour change here.


dzaguri-nv:
RERUN

dzaguri-nv:
RERUN

vadimp-nvidia (approved):

acoifmannvidia (approved):

acoifmannvidia:
RERUN

sholeksandr:
RERUN

dzaguri-nv:
RERUN

dzaguri-nv:
RERUN

PR #2948, opened 2026-09-22
hw-mgmt:scripts:  Refactor Redfish BMC login flow

sholeksandr:
1. Split RedfishClient POST /login JSON handling into
__apply_login_response so curl errors stay separate from token parsing.

2. Restructured BMCAccessor.login() so each fallback step (A-D) is one
explicit action, cache the TPM password once per attempt, and centralize
Pass/Fail flow logging in _report_bmc_login_flow. Only record Z'1 when
factory NOS login succeeds but TPM password reset fails. Guard
restore_tpm_credential in __getattr__ when the method is not implemented.

Bug: 5276528 (Defect 2)


sholeksandr:
RERUN

acoifmannvidia (approved):

sholeksandr:
RERUN

PR #2950, opened 2026-09-22
hw-mgmt: thermal: Fix PWM stuck high after tacho error recovery

sholeksandr:
Issue: After a simulated fan tacho error, PWM correctly rose to the
fan_err.tacho dmin (65%). After restoring the real tacho sysfs, PWM
stayed at 65% and never recovered. Seen with TC 2.0 (q3400 fan error
check, drwr1:[1, 2]).

RC: _validate_rpm kept cached fan_tacho_state while PWM was treated as
unsettled, so the tacho fault never cleared and handle_err held tz_pwm
at
dmin. TC 2.0 uses integer PWM from sysfs, but dmin may still pass float
setpoints (65.0 vs 65). fan set_pwm could extend RPM relax before an
ASIC-not-ready return, leaving pwm_set out of sync with chassis PWM.

Fix: Treat read vs rounded set as settled when |delta| < 1% as integers;
adjust multi-tacho cache handling; normalize set_pwm and defer relax
until ASIC-ready; add offline test for float dmin setpoint.

Bug: 78069


sholeksandr:
RERUN

PR #2953, opened 2026-09-23
hw-mgmt: Fix chipup I2C trace collection and add SPC1 PWM fallback (#…

acoifmannvidia:
…2932)

- Capture chipup I2C trace reliably: dedicated ftrace instance, correct adapter filter, bounded buffer size, and safe rotation.
- SPC1: fall back to 100% PWM via mlxreg MFSC after a warm-reboot mlxsw_minimal chipup failure, since thermal sysfs PWM is missing.
- Fix trace cleanup (stale ftrace instances, leaked lock FDs) and make archive naming collision-safe.

Bug #4966961


sholeksandr (approved):

acoifmannvidia:
RERUN

PR #2956, opened 2026-09-23
recipes-kernel: add Greptile review rules for kernel patches

dzaguri-nv:
Advisory .greptile/ config scoped to recipes-kernel/linux/. Greptile checks the patch header, author, description and status-table registration, and comments on the PR - it does not block the merge.

The two checks worth having: a patch with no Patch_Status_Table row is silently skipped by deploy_kernel_patches.py, and a typo inside take[]/skip[] is ignored rather than rejected. Both ship broken with no error.

Upstream cherry-picks are exempt from the description rules.

Also a repo-root rule rejecting AI attribution - every author, Signed-off-by and Co-Authored-By must be a real person with a real email.


dzaguri-nv:
RERUN

